PR #79, opened 2020-11-19
Added redirect for sound.

syntheticmagus:


vercel.json
```diff
@@ -583,6 +583,7 @@
         { "source": "/how_to/webxr_selected_features", "destination": "/divingDeeper/webXR/WebXRSelectedFeatures" },
         { "source": "/How_To/WebXR_Session_Manager", "destination": "/divingDeeper/webXR/webXRSessionManagers" },
         { "source": "/how_to/webxr_session_manager", "destination": "/divingDeeper/webXR/webXRSessionManagers" },
+        { "source": "/how_to/playing_sounds_and_music", "destination": "/divingDeeper/audio/playingSoundsMusic" },
```
syntheticmagus, line 586: I'm not sure I put this in the right file, but I came across this link in our documentation.

RaananW: That's the right place to add this, can you change the doc as well when you get the chance? Instead of redirecting it will be better to have a clean doc link :-)

RaananW: Decided to take care of all of the wrong internal links now, so all good :-)

RaananW (approved):

PR #104, opened 2020-12-07
Typo in opacity texture code

bestog:


PirateJC (approved):
Looks great thanks for the fix!

PR #120, opened 2020-12-22
Formatting, clarity, and links

richardtallent:
I went through (mostly) the Camera pages and adjusted the text to address markdown, grammar, spelling, brands, and reading clarity. I was trying to understand this content more deeply, so that was a good excuse to try to make some improvements where I found some confusion, and where Grammarly caught issues.

I also added a number of internal links from classes, methods, and properties mentioned in the pages back to the API (usually the first mention, with the following instances just back-ticked).

A number of changes were also automatically made to these same files by Prettier as I saved them, but those changes seem in line with current guidelines, so hopefully they aren't an issue. Prettier can be annoyingly opinionated, but it has done wonders for my code consistency, so I recommend it, especially when working in a team. I did address one issue in the Prettier config file so it won't add three spaces after the "-" of a list item (this also fixed the JS/TS tab spacing to 2, which isn't my personal preference but is more common).

I added a `.markdownlint.json` file to allow that tool to function without giving unnecessary warnings. One thing markdownlint caught that I wasn't sure how to address was that it flagged H1s at the top of the page where there is already a `title` in the YAML header block. It considers this a duplicate page title. For now, I disabled that error in its configuration file, but if the team's intention is to avoid H1s for that reason, it can be easily re-enabled.

I definitely let myself wander too far before I made a commit... sorry about that! I'll do smaller commits in the future.

### DIFF
--- a/.markdownlint.json
+++ b/.markdownlint.json
@@ -1,0 +1,9 @@
+{
+  "line-length": false,
+  "no-inline-html": {
+    "allowed_elements": ["Playground", "Youtube", "kbd"]
+  },
+  "single-title": {
+    "front_matter_title": ""
+  }
+}
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,5 +1,5 @@
 {
-  "trailingComma": "all",
-  "tabWidth": 4,
-  "printWidth": 300
+    "trailingComma": "all",
+    "tabWidth": 2,
+    "printWidth": 300
 }
--- a/content/How_To/WebXR/WebXR_Controllers_Support.md
+++ b/content/How_To/WebXR/WebXR_Controllers_Support.md
@@ -1,6 +1,6 @@
 ---
 title: WebXR Controllers Support
-image: 
+image:
 description: Learn about the robust library of WebXR controllers and input supported in Babylon.js.
 keywords: babylon.js, diving deeper, WebXR, VR, AR, input, controller
 further-reading:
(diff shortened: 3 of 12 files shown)


sebavan (approved):
As there is not only text but also lint changes, I ll leave it to @RaananW to merge it as soon as he ll be back :-)

RaananW:
Hey!

Thanks. I'm mobile right now so i can't test the linter, but i have no problems with having it there.

I want to be sure that every person editing a page using the editor of their choosing will be able to write proper markdown without the need of any external plugins. If that's the case I'll approve and merge it.

The reason for the title and the h1, by the way, is that front matter is technically optional. There are fallbacks on each level (h1  - matter - structure.json) and i do prefer it would stay this way.

richardtallent:
Hey @RaananW!

Yup, the linter is just a VS Code extension I've installed that runs live, it wouldn't be part of the build process and wouldn't impact anyone not already using that extension. All I added was a config file for it, so it has a more nuanced ruleset for the users who do have it installed.

I think I understand you on the H1 vs. YAML thing. The linter by default only warns when a single page has both and H1 and a YAML title -- either one by itself is fine. If what you're saying is that a single page *should* be able to have both, the config I added will allow that. But if you're trying to allow either one but *not* both, that config option can be removed to warn about the duplication.

RaananW:
Thanks for explaining 😊

All sounds good, merging

PR #123, opened 2020-12-27
Add CSG reference and example to Merging Meshes page

richardtallent:
Based on my experience trying to figure out how to merge meshes, I've added a reference to the CSG class and an example for using it to the existing page that covers more traditional mesh-merging.

The other changes on the page were just prettier doing prettier things.


RaananW (approved):
Thanks! CSG fits, i totally agree 😊

PR #178, opened 2021-02-26
inspector showing code was missing

Vinc3r:
I wasn't able to find the `scene.debugLayer.show();` piece of code in the inspector page, so here you have


Vinc3r:
sorry for the ping @deltakosh , but is this PR a problem? no news since 11 days, which is very rare from BJS team 😊

RaananW:
> which is very rare from BJS team 😊

Very rare!

My bad, thanks a lot!!!

Vinc3r:
thanks to you ;)

PR #183, opened 2021-03-09
many fixes, separated by commits

Vinc3r:
1. `How_To/` in images links wasn't work online, I replaced it by `how_to/`
2. Added IBL texture tool link
3. reorder some parts in a more logical way to me, and also remove the twice part about Sandbox `generate .env` button


RaananW:
There are a few build issues, mainly because you changes all `How_To` to `how_to`. These refer to the directory the .md file is located at, and not the URL. Also, the `How_To` url are being correctly forwarded to their new location, but `how_to` URLs simply fail with 404.
The image URLs are the only ones that are served from the `how_to` URL, because this is the actual name of the directory in the public library.

Vinc3r:
wooops ok! i fix this

Vinc3r:
should be fine now, right?

RaananW:
Was it intentional that the png you submitted is empty? Any reason to actually show it in the docs?

Vinc3r:
not intentionnal at all 😱 but I don't understand, on my fork we clearly see the image https://github.com/Vinc3r/BabylonJS-Documentation/blob/master/public/img/how_to/Environment/ibl-texture-tool.png

I will resave the png, maybe there was a kind of corruption

Vinc3r:
I really doesn't understand what happens here... here the image from imgur: https://i.imgur.com/1hP3BjT.png

do you want me to delete this file, and try later?

(sorry about all these issues...)

RaananW:
I have to admit that it is odd. 

image

The commit seems to be deleting the file?

Vinc3r:
i've update my PR so as to get rid of this image, we will see this later 🙃

PR #185, opened 2021-03-10
Correcting typos

DawitJung:
Thank you for your kind document.

I found that the code in the document was different from the code in the playground.
And I guess the code on the playground is correct.
This does not cause errors, but I think it can cause misunderstanding.


RaananW (approved):
Naming the material in any name will still work correctly, so the code is valid. Nevertheless, thank you so much for keeping us consistent :-)

PR #196, opened 2021-03-23
Add docs for ArcRotateCamera.zoomToMouseLocation

lovettchris:
The following optional `ArcRotateCamera` properties can also be handy:

- **zoomToMouseLocation** - if set to `true` will cause mouse wheel to zoom in or out centered on
  current mouse location instead of a fixed camera.target location.  This makes it easy  to explore
  all corners of a large scene.  Setting this means that the mouse wheel input controller will be
  changing the camera.target position during mouse wheel zooming.  When this is `true` the zoom
  operation using the mouse wheel is doing both zooming and a small amount of panning at the same
  time.

- **wheelDeltaPercentage** - if set to non-zero value will cause the zooming amount to be set to a
  percentage of the camera radius.  This means the zoom slows down as you get closer to the target
  object which is nice because it means you can be more precise about camera placement while you
  explore your object up close.

See https://playground.babylonjs.com/#GT8E7R#1 for example.

### DIFF
--- a/content/How_To/camera/camera_introduction.md
+++ b/content/How_To/camera/camera_introduction.md
@@ -91,16 +91,16 @@ Whether using the keyboard, mouse, or touch swipes, left/right directions change
 The following optional `ArcRotateCamera` properties can also be handy:
 
 - **zoomToMouseLocation** - if set to `true` will cause mouse wheel to zoom in or out centered on
-  current mouse location instead of the center of the target object.  This makes it easy  to explore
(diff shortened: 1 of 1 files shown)


PirateJC (approved):

PirateJC:
Looks fantastic! Thanks @lovettchris !!!

lovettchris:
Glad to help, BabylonJS is a very cool project.

kuladeeparun:
Hey @lovettchris! Thanks for this addition. But the playground link isn't working as expected. There's no scene response on scroll up

PR #203, opened 2021-03-31
Adding Gui Demo to Docs

msDestiny14:


PirateJC:
Made a few TINY changes, but looks great!

PR #208, opened 2021-04-10
Allow newer nextjs patchlevels, to fix problems on some systems

egnor:
Fixes required for `yarn install` and `yarn run dev` to work on my system (vanilla Ubuntu 20.04):

- Allow patchlevel updates of nextjs (10.0.x instead of 10.0.5). On my system, there's an issue in nextjs <10.0.7 with some complicated version interaction that causes sharp's install script to fail to find precompiled binaries for libvips; it tries to compile libvips from source, which doesn't work because expectations about certain header files (glib-object.h) don't match the system layout.

- Add an explicit (dev) dependency on the same webpack version nextjs uses, due to issues https://github.com/vercel/next.js/issues/21679 (duped with https://github.com/vercel/next.js/issues/21721). The root cause there is somewhat mysterious to me but without the webpack dependency, the dev-server can't generate pages, complaining that NodeTemplatePlugin is missing.

- The `yarn.lock` changes that result from the above.

Note, even after these changes, `yarn run build` doesn't work for me:

```
Error: Image Optimization using Next.js' default loader is not compatible with `next export`.
  Possible solutions:
    - Use `next start` to run a server, which includes the Image Optimization API.
    - Use any provider which supports Image Optimization (like Vercel).
    - Configure a third-party loader in `next.config.js`.
    - Use the `loader` prop for `next/image`.
  Read more: https://err.sh/next.js/export-image-api
    at /home/egnor/source/BabylonJSDoc/node_modules/next/dist/export/index.js:14:785
    at async Span.traceAsyncFn (/home/egnor/source/BabylonJSDoc/node_modules/next/dist/telemetry/trace/trace.js:5:584)
```

I'm not sure if the HTML export step is supposed to work? (For local editing, `yarn run dev` is all you really need.) It seems like doc.babylonjs.com has a "Powered by Vercel" logo; does that mean Vercel is in fact serving the pages and handling the image optimization?

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,7 +14,7 @@
     "@next/mdx": "^10.0.5",
     "@types/puppeteer": "^5.4.0",
     "mdast-util-toc": "^5.0.4",
-    "next": "10.0.x",
+    "next": "10.0.9",
     "next-mdx-remote": "^2.1.1",
     "node-html-parser": "^2.0.0",
     "prism-react-renderer": "^1.1.1",
--- a/package.json
+++ b/package.json
@@ -14,7 +14,7 @@
     "@next/mdx": "^10.0.5",
     "@types/puppeteer": "^5.4.0",
     "mdast-util-toc": "^5.0.4",
-    "next": "10.0.5",
+    "next": "10.0.x",
     "next-mdx-remote": "^2.1.1",
     "node-html-parser": "^2.0.0",
     "prism-react-renderer": "^1.1.1",
@@ -43,6 +43,7 @@
     "typedoc": "^0.19.2",
     "typescript": "^4.0.5",
     "unified": "^9.2.0",
-    "unist-util-visit": "^2.0.3"
+    "unist-util-visit": "^2.0.3",
+    "webpack": "4.20.x"
   }
 }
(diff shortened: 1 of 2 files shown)


RaananW:
Thank you so much for that!

next 10.0.6 had a serious issue that prevented us (and many other projects) from building correctly. This is why I am not very keen on using a `x` version instead of a fixed version we are sure working. I am updating the packages from time to time, I will make sure to update next to the latest asap.

Regarding the build issues with image optimization, I will put this on my todo list. the issue is with server-sided image optimization not available locally. `start` and `dev` tasks should work without a problem.

RaananW:
reopening this - would you want to set the fixed version that works for you? I will test the build process and merge it then

RaananW:
To eliminate the second error use `yarn run build:local` instead of `build`, this PR added the task - https://github.com/BabylonJS/Documentation/pull/211

egnor:
Happy to sub in anything 10.0.7 or newer! Do you have a preference between 10.0.7 (the first version that works for me) and 10.0.9 (the latest 10.0.x)?

RaananW:
latest fixed version is fine for me :-)

Thanks!!

egnor:
> latest fixed version is fine for me :-)

OK! Changed from 10.0.x to 10.0.9, take another look?

PR #219, opened 2021-04-27
Overhaul "Reflections and Refractions" page (and some side infrastructure/style/content improvements)

egnor:
This turned into a bit of a yak shave, my apologies! I am happy to break this up if that works better for you. Also of course I am super open to feedback on any of this -- I am new here so please make sure I got things right, both in terms of technical accuracy and style.

Content:

Revise the "Skyboxes" and "Reflections and Refractions" pages (and a few references elsewhere) to cover a number of things that were confusing to me, having had them explained. Explain the principles of environment mapping, the differences between static and dynamic environments, the differences between cubemaps and flat textures, and give lots of hopefully simple and clear examples of the various concepts. Explain the different coordinate modes and what they're for, and add a discussion of the reflection transform. Add links to other resources (API docs and explainers).

Infrastructure/Style:

I was fighting to get images to render right (kept having weird spacing, captions didn't work, funny business on resize, etc), and took a deep dive into the image handling logic. I ended up changing it to use ordinary `` (instead of the fancy nextjs `<Image>` component) when width= and height= are not specified -- `<Image>` doesn't do much good in this case, and the runtime Javascript to resize things on the fly was causing a number of problems (validation errors, layout issues, captions were broken, etc). Things seem to work MUCH better now, images are more "normal", and the responsive layout is more functional. I also added comments with my understanding of the logic.

Miscellaneous other changes:

- Added vim swapfiles and emacs backup files to .gitignore
- Added "next-remote-watch" as a dev dependency, and "yarn watch" as a command to run it; it's a very convenient auto-reloader when editing nextjs files
- Improved the handling of document image thumbnails (used in see-also boxes); the "image:" yaml field was broken, and the regexp was insufficiently selective in some cases


RaananW:
Hi, 

Thanks so much for investing the time!

Let's close this PR and divide those into issues and individual content changes. It's very hard to review it this way.

I'll be happy to discuss those changes individually. Maybe even before submitting the code changes. I want to understand what are the actual issues. 

Also, I'm not a big fan of pure html in md files.

egnor:
Fair enough! How do you want to do the discussion? Should I describe the
changes in email, or a forum post?

RaananW:
If you want to submit there here as individual issues we can discuss them here, if you feel more comfortable discussing them on the forum (what we usually prefer) you can do that there. 

As it seems like you have a few issues with the code and a few issues with the comment, a single forum post will probably not be enough. Start by submitting the issues here, then we have a point of reference.

egnor:
OK, split the mega-PR up into a bunch of hopefully more digestible units :-)

PR #225, opened 2021-05-06
Add vim and emacs swap/backup files to .gitignore

egnor:
These are useful for us Unix command line weirdos. 🤓


RaananW (approved):
We support all weirdos!

PR #228, opened 2021-05-06
Add next-remote-watch to devDependencies (and set up "yarn watch" to run it)

egnor:
next-remote-watch (https://github.com/hashicorp/next-remote-watch) causes pages to automatically reload when content inputs (such as .md files) are changed. (The "yarn watch" command is set up to point at the content/ directory.)

I find this handy for editing, but it's a nicety for sure and completely reasonable to leave out if it seems frivolous.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,8 @@
     "dev": "next dev",
     "build": "next build && next export",
     "build:local": "next build",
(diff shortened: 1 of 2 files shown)


RaananW (approved):
LGTM!

PR #238, opened 2021-05-09
Added a Playground example for WebXR controllers input

flushpot1125:
Added a Playground example for WebXR section according to the post of forum.

https://forum.babylonjs.com/t/webxr-based-vr-controller-input/20497  

Hope this will be useful.


RaananW (approved):
Thanks a lot!

PR #247, opened 2021-05-26
add WebXRDomOverlay information

brianzinn:
please wait for next release (alpha.23) when available on NPM before merging.


RaananW:
I think this belongs in the AR features more than here. What do you think?

brianzinn:
> I think this belongs in the AR features more than here. What do you think?

Absolutely - I definitely put in wrong place. Will move it to https://doc.babylonjs.com/divingDeeper/webXR/webXRARFeatures

RaananW (approved):

PR #289, opened 2021-07-25
add doc about cesiumJS

Hypnosss:


RaananW (approved):
Thanks a lot for that!

PR #294, opened 2021-07-29
Updates to height-map documentation

sokorototo:
Cleared some spelling and grammar mistakes, I think. Reduced the amount of information delivered for a more streamlined learning experience. Changed code examples to use cleaner more readable code. Removed `var` in favor of `const`. Changed structure of the document a little bit.

### DIFF
--- a/content/How_To/mesh/creation/set/height_map.md
+++ b/content/How_To/mesh/creation/set/height_map.md
@@ -25,8 +25,7 @@ Understanding height maps is the main objective of this tutorial. A height map i
 
 ![HeightMap3](/img/how_to/HeightMap/worldHeightMap.jpg)
 
-> To help you generate height-map textures (like the one above), you can use software such as “Terragen”, or ”Picogen”.
->
+> To help you generate height-map textures (like the one above), you can use software such as [Terragen](https://planetside.co.uk/free-downloads/terragen-4-free-download/), or [Picogen ](https://picogen.org/downloads.php.html)(free).
 
 ### Javascript code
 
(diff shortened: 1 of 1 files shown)


RaananW:
@PirateJC - would be great to get your review here.

Just a question from me - why have you decided to remove so much of the content?

sokorototo:
Because as I was passing through the docs I noticed the information was kind of unfocused. So I made a PR and made some adjustments. I tried my best to hit all the important information. Maybe you can tell me what important points I missed in specific. I also added some information, from the API docs.

sebavan:
Agree that it sounds weird to lose some info

sokorototo:
What important info has been lost?

sebavan:
Like the camera tips code for instance. I ll let @PirateJC as he is way more aware than me about the doc progression

sokorototo:
I think that it was unrelated to the topic at hand, and should most likely be in another chapter. A "tips and tricks" chapter for example. Things like preventing clipping into the height map could be discussed.

PirateJC:
Just read through this new version and I honestly think it's a fantastic improvement! It reads more clearly and helps build the information better than in the previous iteration. Fantastic work @sokorototo! You clearly have a gift for this!

I tend to agree I actually don't think the camera piece is a critical part of this particular doc, more of an "additional topic."

My only ask before we merge is to add links to when you mention Terragen and Picogen. @sokorototo would you mind doing the honors for that?

Other than that I think it's a nice rewrite of the doc. Again great work!

sokorototo:
Yeah sure

sokorototo:
I have now added download links to Terragen's and Picogen's download pages. Thanks for all the feedback guys

PirateJC (approved):
Awesome! Approved! Thank you so much for doing this awesome pass! Fantastic work!

PR #348, opened 2021-10-11
GUI Editor documentation edits

msDestiny14:


PirateJC (changes requested):
It looks like you forgot to include this image in the PR:

changeName.jpg

Can you add it please?

msDestiny14:
ah I see what I did. We can remove the extra line with the picture since the picture above had both parts in it. I will change the text. 

Edit: Actually on second thought I will include the picture. :)

PirateJC (approved):

PR #368, opened 2021-10-28
Documentation for Vertex Animation Textures

brunobg:
This is the documentation for https://github.com/BabylonJS/Babylon.js/pull/11317.

### DIFF
--- a/content/How_To/animation/baked_texture_animations.md
+++ b/content/How_To/animation/baked_texture_animations.md
@@ -103,7 +103,7 @@ Then you set the parameters for the thin instances:
 
 Here's an example:
 
-<Playground id="#CP2RN9#18" title="Vertex Texture Animations on thin instances" description="An example of playing VATs on thin instances."/>
+<Playground id="#CP2RN9#20" title="Vertex Texture Animations on thin instances" description="An example of playing VATs on thin instances."/>
 
 ## VATs with SPS
 
--- a/content/How_To/animation/baked_texture_animations.md
+++ b/content/How_To/animation/baked_texture_animations.md
@@ -20,49 +20,48 @@ The `VertexAnimationBaker` class generates a texture for you given the animation
 
 ```js
 let baker = null, mesh = null;
+const animationRanges = [{from: 1, to: 20, name: "My animation"}];
 BABYLON.SceneLoader.ImportMeshAsync(
-    "",
-    "https://raw.githubusercontent.com/RaggarDK/Baby/baby/",
-    "arr.babylon",
-    scene,
-    undefined
+  "",
+  "https://raw.githubusercontent.com/RaggarDK/Baby/baby/",
+  "arr.babylon",
+  scene,
+  undefined
 ).then((importResult) => {
-    mesh = importResult.meshes[0];
-    baker = new BABYLON.VertexAnimationBaker(scene, mesh);
-    // you can slice the animation here with several animation ranges.
-    return baker.bakeVertexData([{from: 1, to: 20, name: "My animation"}]);
+  mesh = importResult.meshes[0];
+  // create the baker helper, so we can generate the texture
+  baker = new BABYLON.VertexAnimationBaker(scene, mesh);
+  // you can slice the animation here with several animation ranges.
+  return baker.bakeVertexData(ranges);
 }).then((vertexData) => {
-    const vertexTexture = baker.textureFromBakedVertexData(vertexData);
-    // your texture is ready to use, see below
-});
-``` 
+  // we got the vertex data. create the texture from it:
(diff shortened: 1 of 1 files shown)


sebavan:
Let s merge it when we ll merge https://github.com/BabylonJS/Babylon.js/pull/11317 or ppl might find it before it is available :-) but it looks great

content/How_To/animation/baked_texture_animations.md
```diff
@@ -0,0 +1,176 @@
...
+keywords: diving deeper, animation, advanced
+further-reading:
+video-overview:
+video-content:
+---
+
+## Baked vertex animations
+
+Animations are computed by the CPU, applying the bone effects to the mesh. This is reasonably slow and for several animated objects can be a bottleneck. One way to optimize this is pre-computing (or baking) the animations, storing them into a texture (usually called Vertex Animation Textures, or VAT) and using it on the vertex shaders . This frees the CPU, with the trade-off that you need to perform this initial baking step (which can be done at development time), add a new texture file to your downloads and consume more GPU memory. This trade-off is usually quite good, since the CPU tends to be the bottleneck.
+
```
Popov72, line 14: In the trade-off / limitation: the animations may not be as smooth as when not using BVA because there is no interpolation between frames. The "smoothness" will depend on the number of frames of the animation.

content/How_To/animation/baked_texture_animations.md
```diff
@@ -0,0 +1,176 @@
...
+## VATs for instances
+
+As explained in [How To Use Instances](/divingDeeper/mesh/copies/instances), instances are an excellent way to use hardware accelerated rendering to draw a huge number of identical meshes. VATs can be used further to handle the animations efficiently.
+
+<Playground id="#CP2RN9#13" title="Vertex Texture Animations on instances" description="An example of playing VATs on instances."/>
+
+## VATs for thin instances
+
+VATs can also be used with [thin instances](/divingDeeper/mesh/copies/thinInstances). In this case you need to register a buffer to set the animation parameters for each instance.
+
```
Popov72, line 76: > In this case you need to register a buffer to set the animation parameters for each instance.

This is also true for the "standard instances" case (where you use `registerInstancedBuffer`) when in the thin instance case you use `thinInstanceSetBuffer`.

content/How_To/animation/baked_texture_animations.md
```diff
@@ -0,0 +1,176 @@
...
+Here's an example for a single mesh:
+
+<Playground id="#CP2RN9#16" title="Vertex Texture Animations" description="An example of playing a vertex texture animation."/>
+
+
+## VATs for instances
+
+As explained in [How To Use Instances](/divingDeeper/mesh/copies/instances), instances are an excellent way to use hardware accelerated rendering to draw a huge number of identical meshes. VATs can be used further to handle the animations efficiently.
+
+<Playground id="#CP2RN9#13" title="Vertex Texture Animations on instances" description="An example of playing VATs on instances."/>
```
Popov72, line 71: This PG does not use the new manager, it still uses the old way through `material.bakedVertexAnimationMap`. I think you want to use the `#18` version of the PG instead.

content/How_To/animation/baked_texture_animations.md
```diff
@@ -0,0 +1,176 @@
...
+## VATs for thin instances
+
+VATs can also be used with [thin instances](/divingDeeper/mesh/copies/thinInstances). In this case you need to register a buffer to set the animation parameters for each instance.
+
+```js
+  // create the instanced buffer
+  mesh.registerInstancedBuffer("bakedVertexAnimationSettingsInstanced", 4);
+  // set it for the base mesh
+  mesh.instancedBuffers.bakedVertexAnimationSettingsInstanced = new BABYLON.Vector4(0, 0, 0, 0);
+```
```
Popov72, line 82: This code is for the "standard instances" case => my mistake, it should not be in the thin instance PG.

content/How_To/animation/baked_texture_animations.md
```diff
@@ -0,0 +1,176 @@
...
+    animParameters.set(params.asArray(), i * 4);
+  }
+
+  // update the mesh with all settings
+  mesh.thinInstanceSetBuffer("bakedVertexAnimationSettingsInstanced", animParameters, 4);
+```
+
+Here's an example:
+
+<Playground id="#CP2RN9#18" title="Vertex Texture Animations on thin instances" description="An example of playing VATs on thin instances."/>
```
Popov72: The right version is `#17` for the thin instance case. However, as there are some wrong lines in this PG, here's one stripped from those lines: https://playground.babylonjs.com/#CP2RN9#20

PR #369, opened 2021-10-30
Update "Using ShaderMaterial" documentation

BlakeOne:
Small update to the "Using ShaderMaterial" documentation to add the samplers array to the example code snippet and change the corresponding line of text to reflect that the samplers array must be used for texture uniforms rather than the uniforms array.

Forum: https://forum.babylonjs.com/t/shader-texture-changing-problem/9640/14


RaananW (approved):

RaananW:
Thanks!!

PR #371, opened 2021-11-01
Update Gui.md

BlakeOne:
Add a note about manual resizing to the adaptive scaling section of the GUI documentation.

Forum: https://forum.babylonjs.com/t/introducing-the-gui-editor-alpha/24578/80

### DIFF
--- a/content/How_To/GUI/Gui.md
+++ b/content/How_To/GUI/Gui.md
@@ -208,6 +208,8 @@ To define this resolution, just set `myAdvancedDynamicTexture.idealWidth = 600` 
 
 If both are set, the idealWidth will be used.
 
+If neither is set, you will be responsible for resizing the UI controls yourself (this can useful if you need greater control over the resizing strategy used).
+
 When ideal resolution is set, all values expressed **in pixels** are considered relatively to this resolution and scaled accordingly to match the current resolution.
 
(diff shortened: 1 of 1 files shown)


BlakeOne:
Or maybe it should say "rescaling" instead of "resizing"? :)

msDestiny14:
I don't know if you need the parenthesis but just my opinion.

BlakeOne:
I went back and forth a bit and got it into my head that it sounded friendlier with the parenthesis somehow, LOL. Here it is without them thou. :)

The grammar checker said you'll would be more friendly than you will - I didn't think it sounded as good, but they got into my head, LOL. ;)

msDestiny14:
Also maybe I'm being very nitty on the wording but what what does this mean? "resizing strategy used". If I had no idea about anything Babylon what would this mean to me. I guess the biggest advice for documentation is assuming the reader may not have knowledge.

BlakeOne:
To me examples of resizing strategies would be "scale by min dimension", "scale by width", etc. It's whatever strategy you want to use for resizing.  It's more of a general-purpose CG term that I've read and used before thou, not Babylon specific per se...

I'm def not married to it thou. :)

BlakeOne:
Here it is with just the one, simpler line. :)

sebavan (approved):

PR #372, opened 2021-11-02
HammerJS Input

RolandCsibrei:
Hi @RaananW!
Here is the PR for the HammerJS camera input.

### DIFF
--- a/content/How_To/camera/Customizing_Camera_Inputs.md
+++ b/content/How_To/camera/Customizing_Camera_Inputs.md
@@ -154,7 +154,7 @@ The example contains `utils\ArcRotateCameraHammerJsInput.ts` which implements `I
 import "hammerjs";
 ```
 
-To use the new input you add it to your `camera.inputs` after you have create the camera. To avoid one input fighting the other remove the `ArcRotateCameraPointersInput` from `camera.inputs`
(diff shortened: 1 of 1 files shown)


content/How_To/camera/Customizing_Camera_Inputs.md
```diff
@@ -136,6 +136,45 @@ detachControl(element);
...
+
+```javascript
+
+    // remove mouse input
+    camera.inputs.removeByType('ArcRotateCameraPointersInput')
+
+    // add hammer js input
+    const hammerJsInput = new ArcRotateCameraHammerJsInput()
+    // now you can set the parameters you like
+    hammerJsInput.xPanningSensitivity = ...
```
RaananW: as this will usually be copied, it would be great to set here a default value that makes sense in a general scenario

content/How_To/camera/Customizing_Camera_Inputs.md
```diff
@@ -136,6 +136,45 @@ detachControl(element);
...
+    hammerJsInput.zoomSensitivity = ...
+    .
```
RaananW: Same concept - maybe remove those characters and add a comment saying something like "// further configuration the input behavior..." or something similar.

RaananW (approved):
Thank you! This is great. 
Just 2 nitpick comments :-)

PR #398, opened 2021-11-30
Dev Stories: Playground to Production - Fruit Fallin'

syntheticmagus:
Draft concept for a potential new "Dev Stories" section, bootstrapping with a dev story for _Fruit Fallin'_.


content/dev_stories/playground_to_production_fruit_falling.md
```diff
@@ -0,0 +1,677 @@
...
+more complex experiences, commercial scenarios... At a high level, the 
+procedure described above can be used to take any Babylon.js idea from 
+Playground to production using the Babylon.js Template Repositories, NPM, 
+and Ionic.
+
+For a less high-level look, let's take a more step-by-step journey through 
+the process.
+
+## A More Step-by-Step Journey Through the Process
+
```
carolhmj, line 80: I think a first step titled something like "Develop your idea/prototype" on the Playground, that links to the Fruit Fallin Playground and also links to the Playground doc page would be cool!

content/dev_stories/playground_to_production_fruit_falling.md
```diff
@@ -0,0 +1,677 @@
...
+[_Fruit Fallin'_](https://syntheticmagus.github.io/fruit-falling-source/), 
+a Babylon.js app inspired by browser games from the heyday of Flash. While 
+I happened to be making a simple game, this same process can be used for 
+more complex experiences, commercial scenarios... At a high level, the 
+procedure described above can be used to take any Babylon.js idea from 
+Playground to production using the Babylon.js Template Repositories, NPM, 
+and Ionic.
+
+For a less high-level look, let's take a more step-by-step journey through 
+the process.
```
carolhmj, line 77: Maybe linking the forum here so anyone following the steps has on their mind that they can head to the forum for help?

content/dev_stories/playground_to_production_fruit_falling.md
```diff
@@ -0,0 +1,677 @@
...
+    I clicked the "Use this template" button to start a new repo.
+
+    !["Use this template" button](/img/devVignettes/fruitFalling/01_use_package_template.jpg!500)
+1.  I picked my repo name, choose to make it public (for non-open source, 
+    I would have chosen private), then created the repository.
+
+    ![Picking repo options](/img/devVignettes/fruitFalling/02_choose_options.jpg!500)
+1.  In a Git-enabled terminal (in my case, Git Bash on Windows), I 
+    navigated to where I wanted my new repo to be and created my local 
+    clone.
```
carolhmj: Maybe linking this doc to help users with less git experience? https://git-scm.com/book/en/v2/Git-Basics-Getting-a-Git-Repository

content/dev_stories/playground_to_production_fruit_falling.md
```diff
@@ -0,0 +1,677 @@
...
+The `Sprite` I used while learning how to use sprites is the one provided
+as part of Babylon's default assets. While I knew all along I wanted to
+replace that, I first wanted to use it to bootstrap and test my approach
+to asset hosting for development.
+
+1.  Repeating essentially the steps from the 
+    [prior section](#creating-an-asset-host-repo) 
+    for creating a new repository from a template, I made a new asset host 
+    from the 
+    [Asset Host Template](https://github.com/BabylonJS/npm-package-template),
```
carolhmj: Here should be https://github.com/BabylonJS/asset-host-template?

content/dev_stories/playground_to_production_fruit_falling.md
```diff
@@ -0,0 +1,677 @@
...
+    the the `SPRITE_URL` assignment line to
+
+    ```ts
+    const SPRITE_URL: string = "http://127.0.0.1:8181/player.jpg";
+    ```
+
+    save the file, then wait for the test app to reload.
+1.  Of course, once that reload happened, I realized I wouldn't be able
+    to see the results because the sprite from my asset host looked 
+    identical to the one pulled from the Internet. To make extra sure,
```
carolhmj, line 237: I think here, to be a little more direct, it would be better to recommend users to download a different asset from the start?

content/dev_stories/playground_to_production_fruit_falling.md
```diff
@@ -0,0 +1,677 @@
...
+
+    This is another step I recommend for moving your new repo away from 
+    Playground practices and into production. In order to be able to 
+    support code copy-pasted out of the Playground, the NPM Package
+    Template's `app_package` by default includes a *lot* of dependencies,
+    only a subset of which are likely to be used by any particular 
+    codebase. Fortunately, it's easy to get rid of the unnecessary imports
+    here; all I had to do was open `app_package/package.json` and delete
+    the references to any NPM packages I wasn't using--which, in this case,
+    was most of them.
```
carolhmj, line 297: Might be worth linking to the tree shaking doc? https://doc.babylonjs.com/divingDeeper/developWithBjs/treeShaking

content/dev_stories/playground_to_production_fruit_falling.md
```diff
@@ -0,0 +1,677 @@
...
+
+    Most interactive apps have multiple "modes" of operation, and *Fruit 
+    Fallin'* in particular has two: the game and the 
+    title/menu/options/credits. While there's no real rule for how best
+    to separate the concerns of the different modes, I found it useful to 
+    have them as separate Babylon `Scene` objects. Specifically, I made new
+    `Scene` subclasses with the behavior I wanted for the different "modes";
+    that made it easy for me to transition between modes by simply 
+    disposing one kind of scene and creating a new instance of the other, 
+    thus keeping the logic and resources clean and separated.
```
carolhmj, line 331: The Scene subclassing approach is a great idea! I think showing parts of the code of one of these subclasses to exemplify this approach could be nice, as I expect less people are going to actually look at the codebase by itself, so we'd want to have as much upfront information as possible here. Another option would be linking the classes here like the fixed framerate observable example below.

content/dev_stories/playground_to_production_fruit_falling.md
```diff
@@ -0,0 +1,677 @@
...
+    experience has to do to integrate it is just tell `app_package`
+    the correct URLs to use.
+    
+    ![index.js asset location code](/img/devVignettes/fruitFalling/21_asset_location_code.jpg!500)
+
+I should also briefly mention how I created the assets for the game. 
+Again, this is not an implementation overview; but for those who are 
+curious, here are the various tools I used to make *Fruit Fallin'*.
+
+1.  **Art:** I drew all the backgrounds, the buttons, and all the
```
carolhmj, line 390: Pretty cool section, even shows how simple is to create assets! Linking https://opengameart.org/ as an option here could be cool too.

carolhmj (commented):
Excellent article! Hopefully will help quickstart a lot of Babylon developer careers <3

content/dev_stories/playground_to_production_fruit_falling.md
syntheticmagus: Is tree shaking involved in this? My recollection is that my build size didn't actually change when I did this, which I thought made sense because the tree shaking was causing things I wasn't using to get ignored anyway. Thus, removing these dependencies didn't have any effect on tree shaking, it just reduced the unnecessary overhead of the dev repo itself. @RaananW, have I understood that correctly? If so, I feel like bringing up tree shaking here might be a complicating detour.

content/dev_stories/playground_to_production_fruit_falling.md
syntheticmagus: I ended up linking it a little earlier, even closer to the top.

content/dev_stories/playground_to_production_fruit_falling.md
syntheticmagus: I see what you mean, but I'm really trying to separate concerns between the major steps, so I don't want to talk about changing the app's behavior/visage while bringing it into a dev repo. The thing this is trying to emphasize is that you can get your Playground working as-is in the dev repo, after which you can iterate on it. Suggesting using a different asset probably wouldn't be _too_ big a deviation from that, but it would also remove the opportunity to bring up these common asset-serving pitfalls (wrong names, browser caching, etc.), which is what this step really exists to mention. I've confused myself quite a few times with names and browser-cached assets, so I specifically wanted to show that it's worth being mindful of things like that and knowing how to overcome them.

content/dev_stories/playground_to_production_fruit_falling.md
syntheticmagus: Linking the files is a great idea, I'll definitely do that. Ultimately, I want to revisit practices like this more thoroughly in later Stories when there's more of a coding focus, as going too deep on practices like this in this Story would make it even longer and distract from the "Playground to Production" focus. The linking to the source strikes a really good balance.

content/dev_stories/playground_to_production_fruit_falling.md
syntheticmagus: I didn't use that here as I created all the assets myself, but we should definitely find a way to work that into a future Story and call it out there!

content/dev_stories/playground_to_production_fruit_falling.md
syntheticmagus: Done. I don't think it's a separate section or step -- this Story is a journey *away* from the Playground, so I don't want to spend time showing things *in* the Playground (especially since we already have tons of docs showing that) -- but I think linking the Playground prototype in question makes a lot of sense.

content/dev_stories/playground_to_production_fruit_falling.md
RaananW: @syntheticmagus - I think you are right, it would over-complicate things to mention it. 
If you didn't use those imports before and just imported them, they would have been included in the package, but I can only assume the framework itself needs them (or you need them somewhere else in the code) and they are imported anyhow. 

> tree shaking was causing things I wasn't using to get ignored anyway

in an ideal world this is true, but in babylon, since side effects are turned on, importing already means "i need it" because there might be side effects in this class. So no packer will remove this from your packaged build.

content/dev_stories/playground_to_production_fruit_falling.md
carolhmj: That makes sense!

content/dev_stories/playground_to_production_fruit_falling.md
carolhmj: Ooooh, ok, I thought that could explain why taking the unnecessary imports out is good, but it would be a detour.

PirateJC (approved):

PR #411, opened 2021-12-10
ACEDoc

PirateJC:


PirateJC:
@PatrickRyanMS and @deltakosh...this is a first pass of the ACE Doc. Please have a look through and let me know what you think.

Thanks!

deltakosh:
Excellent!

Some feedback:
- We need to link to ACE from regular animation docs
- My video is a bit awkward in a sense that it is dropped with no context :)
- We could mention the small differences happening when using Animation groups instead of direct animation

PR #412, opened 2021-12-13
Update Blender.md

abow:
Removed Tower of Babel


abow:
https://forum.babylonjs.com/t/tower-of-babel-blender-exporter/25996/3

PR #431, opened 2022-01-05
Add a warning about physics to camera collision

carolhmj:
Because of this topic https://forum.babylonjs.com/t/combining-camera-controller-with-cannon-physics/26379/2 I thought it would be nice to add a warning so users don't try to mix mesh collisons and the physics engine.


content/How_To/camera/cameras_mesh_collisions_and_gravity.md
```diff
@@ -48,7 +48,10 @@ scene.gravity = new BABYLON.Vector3(0, earthGravity / assumedFramesPerSecond, 0)
...
 
-If you need a more accurate representation of gravitational (or other) forces, you can add your own physics engine. See [Add Your Own Physics Engine](/divingDeeper/physics/addPhysicsEngine).
+If you need a more accurate representation of gravitational (or other) forces, you can use the physics engines [integrated with Babylon](/divingDeeper/physics/usingPhysicsEngine), or [add your own](/divingDeeper/physics/addPhysicsEngine).
+
+#### A Warning
+Mixing mesh collisions and the physics engine will lead to unexpected behavior.
```
RaananW: "under certain conditions". Or better - Adding both physics impostors and setting collision enabled to the same object might lead to unexpected behavior

carolhmj: Thanks, updated!

RaananW (approved):

PR #467, opened 2022-02-16
Fix missing title in 'Transparent Rendering' page

mysunnytime:
The 'Alpha Tested Meshes' title is missing in the document. This is to fix the title error.

image


RaananW (approved):
Good catch :-)

PR #469, opened 2022-02-17
Modify Vite ES6 example to import EnvironmentHelper directly

kaliatech:
Minor change to vite/ES6 WebXR example to import `EnvironmentHelper` instead of relying on side effects from `sceneHelpers`.

I assume a larger goal is to remove side effects imports as much as possible for ES6 and tree shaking.  The `sceneHelpers` was pulling in a number of other modules that are not required for a simple WebXR setup.

Resulting build size decreased ~80kb  (unzipped).


RaananW (approved):

RaananW:
Thanks!!!

PR #471, opened 2022-02-21
Update In-Browser_Mesh_Simplification.md

carolhmj:
Added an example of how to get the simplified meshes directly, updated an example with the old CreateGround function, some corrections


sebavan (approved):
GG

PR #482, opened 2022-03-08
Fix Erosion WebGPU demo

sebavan:
Removing WGSL ignore which looks deprecated between https://playground.babylonjs.com/#C90R62#10 and https://playground.babylonjs.com/#C90R62#12


Popov72 (approved):
I had corrected the validation test but forgot about this doc page.

Thanks!

PR #484, opened 2022-03-10
Added LensRenderingPipeline distortion note on pointer precision

dested:
I wasn't sure how much to stress this point, since the reality is it caused me months of confusion, but I erred on the lighter side.

The corresponding source PR is here: https://github.com/BabylonJS/Babylon.js/pull/12135


PirateJC (approved):
Thank you for the update @dested. I agree the lighter approach is probably best for now. We can always expand it in the future if we need to provide deeper explanation. But this seems great for now.

PR #485, opened 2022-03-10
VideoDome doc improvements

kaliatech:
Changes
 * Minor updates and additions to "divingDeeper/materials/using/videoTexture"
 * Minor updates and additions to "divingDeeper/environment/360VideoDome"
 * Adding missing PG preview image for a previous PG example in videoDome
 
Related forum post:
 * https://forum.babylonjs.com/t/videodome-clicktoplay-and-dispose-maybe-not-as-expected/28271

Note
 * autoPlay=false is problematic on Safari/iOS.  I attempted to provide simple code/docs that would work on all browsers/devices, but I ran out of time.  I believe there are some underlying issues when autoPlay=false, possibly related to video preload and/or loading events, such I couldn't come up with a simple PG that would show manual play of VideoDome on iOS/Safari reliably.


PirateJC (approved):
@kaliatech these are awesome updates! Thank you so much for taking the time to make them! You rock!

PR #500, opened 2022-03-20
Add missing or incorrect shadow troubleshooting info

AJamesPhillips:
Copied from this question: https://forum.babylonjs.com/t/how-to-fix-shadow-shape/28580

### DIFF
--- a/content/How_To/light/Shadows.md
+++ b/content/How_To/light/Shadows.md
@@ -392,6 +392,14 @@ So, go back and change the bias to its maximum before seeing peter panning and t
 
 Your shadows are now soft without acnea or peter panning.
 
+### Missing or incorrect shadows
+
+Usually this indicates the resolution of the shadow is too low.  It could be that your directional light is too far away from the object, the resolution of shadow texture is too low, or the model is too small.
+
+If the light is far away, you can try to move your light closer to the object.
+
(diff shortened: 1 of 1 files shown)


AJamesPhillips:
Actually we should either:

1. put a link to the question so that if the image URL changes then someone can more easily find the original image and optionally update it to fix it.
2. or put this / another image into this repo

RaananW:
Thank you so much for that :-)

This is one of many reasons as to why you don't see your shadows. Whereas this is a great start, I don't want people to think this is the only thing they can try if their shadows don't work. This should probably expand a little more. Maybe instead of a very abstract title we can add a sentence saying it is possible to try if you are having difficulties with the shadows as part of the text before the newly added section?

AJamesPhillips:
> Thank you so much for that :-)

My pleasure :)

> This is one of many reasons as to why you don't see your shadows. Whereas this is a great start, I don't want people to think this is the only thing they can try if their shadows don't work.
> This should probably expand a little more.

Yes sorry I do not know of any of the other approaches to fixing missing shadows.

> Maybe instead of a very abstract title we can add a sentence saying it is possible to try if you are having difficulties with the shadows as part of the text before the newly added section?

Sorry I do not know how to improve it, could you do this please?  Or I will be happy to update the PR with some text you write here.

RaananW:
Closing this for now, but I totally understand the need for that. I'll create an issue and see how we tackle this.

PR #503, opened 2022-03-22
Add example of faceColors and material

AJamesPhillips:


RaananW:
I don't think this works quite like you intended it to. A material can be applied to any mesh, even with face colors defined. Is that what you meant to say?

AJamesPhillips:
> I don't think this works quite like you intended it to. A material can be applied to any mesh, even with face colors defined. Is that what you meant to say?

Yes.  What I meant to say, and show, was that faceColors and material are composed together.  This enables you to use faceColors at the same time as using a material with specularColor... as a newcommer to BabylonJS I wanted to use faceColors but could not find any documentation on how I could change the specularColor and only through experimentation did I find I could use a material.  I was trying to remove any similar confusion or uncertainty someone else might have by allowing them to answer the question: "Can I / How can I change the specularColor of a mesh which has been initialised with faceColors?"

Please improve the description as you see fit or suggest something and I'll happily update this PR.

PirateJC:
We are redoing the organizational structure of our docs so closing this PR as a result:

Here's the new PR with the bigger organizational overhaul:
https://github.com/BabylonJS/Documentation/pull/622

PR #525, opened 2022-04-12
Multiple platforms landing page first draft.

syntheticmagus:


PirateJC (approved):

syntheticmagus:
Adding @bghgary and @thomlucc, per offline suggestion. Thanks for taking a look!

content/landing_pages/divingDeeperPlatforms.md
```diff
@@ -0,0 +1,120 @@
...
+description: Learn all using Babylon.js on multiple platforms.
+keywords: diving deeper, cross-platform, platform
+further-reading:
+video-overview:
+video-content:
+---
+
+[![Web app](/img/flatIcons/Web-256.png!128)](#web-app) &nbsp &nbsp [![Mobile app](/img/flatIcons/Ipad-256.png!128)](#mobile-app) &nbsp &nbsp [![Desktop app](/img/flatIcons/Computer-Desktop-256.png!128)](#desktop-app)
+
+(Royalty-free icons from [Flaticons.net](https://flaticons.net/).)
```
bghgary: I'm not sure you really need this. The docs automatically generate a TOC, so this is duplicate info? Also, it's hard to understand what is happening with the icons.

image

bghgary (approved):
Looks good except for the TOC

content/landing_pages/divingDeeperPlatforms.md
syntheticmagus: Funky. Locally, the page renders to have the icons look like this:

image

Not sure why Vercel shows it different from this, but the purpose of the icons is just to visually break up what would otherwise be an immediate wall of text as soon as the page loads. Not a make-or-break issue, but when they render correctly I do think they make the page more friendly. Spoke to Jason about it, and we think the best plan is to leave them in for now and, if they also render wrong in the production site, ask Raanan to help us figure that out when he gets back.

bghgary: > the purpose of the icons is just to visually break up what would otherwise be an immediate wall of text as soon as the page loads

Maybe don't make them links then. It might be just me, but I'm not sure what the icons mean by looking at them. It's also weird that the first piece of non-TOC text is "Royalty-free icons ...".

PirateJC: @RaananW to take a look

RaananW: can you merge from master and publish a new version? I will be able to check the preview and the way the page was built.

Having said that, I agree with Gary - it feels weird that those icons are there, especially since it's the only page that has these kinds of links to the content.

PirateJC: I talked to Raanan, we're going to remove the icons for now and revisit later.

@syntheticmagus - Can you please remove these?

PR #529, opened 2022-04-14
Add all current packages

BabylonJSGuide:
Ref https://forum.babylonjs.com/t/new-build-system-in-our-main-repository/28483/52


BabylonJSGuide:
This would overwrite a new version of the framework version page that @RaananW has submitted a PR for.

PR #555, opened 2022-05-09
Adding floating-origin how-to

imerso:
Draft floating-origin how-to. Might need some tweaking.


content/How_To/floating_origin/floating_origin.md
```diff
@@ -0,0 +1,126 @@
...
+title: Floating Origin (Huge Scenes Support)
+image: 
+description: Learn how to manage huge scenes using floating-origin trick
+keywords: diving deeper, huge spaces, floating-origin
+further-reading:
+video-overview:
+video-content:
+---
+
+## HOW FLOATING-ORIGIN WORKS
```
RaananW: Please keep it to lower case (or, well, not all-caps :-))
```suggestion
## How floating-origin works
```

content/How_To/floating_origin/floating_origin.md
```diff
@@ -0,0 +1,126 @@
...
+Camera Position: (10000000, 0, 10000500)  
+Offset: (0, 0, -500)
+
+The offset is small enough to be absolutely precise even with only 32 bits floating-point. The GPU is very happy with that
+and we don't see any jittering.
+
+We just need to set the object at (0, 0, -500) and it will have the same visual effect as if we used their real coordinates,
+but with no jittering.
+
+## FLOATING-ORIGIN EXAMPLE
```
RaananW: Same here
```suggestion
## Floating-origin examples
```

content/How_To/floating_origin/floating_origin.md
```diff
@@ -0,0 +1,126 @@
...
+
+OriginCamera is a special camera which has a separate position control stored as doublepos (and its target as doubletgt). 
+You must stop using position and target from camera, and use their double precision counterparts doublepos and doubletgt.
+
+All objects from scene must then be parented to a Entity instance. Entity also has doublepos property which is double precision coordinate. 
+You must use its doublepos to set object position instead of position directly.
+
+So, let's say that we want a sphere with double precision:
+
+```
```
RaananW: Adding "javascript" after those will make the code format-able in IDEs and will allow the MD parser to work better (i.e. language-specific)

```suggestion
```javascript
```

RaananW (changes requested):
The change requests are mainly already marked as suggestions (but you can change them to whatever fits better!)
I also have a question regarding the doc's position in our docs. Is that the right place for it? Is there no section it fits better? (maybe @PirateJC can comment on that as well).

Otherwise, LGTM!! Thank you so much :-)

content/How_To/floating_origin/floating_origin.md
imerso: Probably in the wrong place, also there are links to my github pages which is probably wrong as well.

Not sure what would be the best to do, maybe those two little classes could be a package hosted somewhere else and the demo made simpler.

If you instruct me what is the best I'll do the required changes. Thanks.

RaananW:
Oh, I think that, if possible, a playground or two would be wonderful. No need to host an extensive demo somewhere, as long as you can show the basic feature.

RaananW (approved):

RaananW:
Just wanted to say - the example is great! but I would make the camera's speed much higher (i.e. - this: https://playground.babylonjs.com/#LHI514#3) otherwise it will take a long time moving around in the scene.

Thanks again, this is awesome :-)

imerso:
Updated with the faster intial speed link. =)

RaananW:
> Updated with the faster intial speed link. =)

did you push the change? i'll merge right after

imerso:
Oops, thought I had -- pushed now.

PR #483, opened 2022-03-09
Possible new organization for Docs

BabylonJSGuide:
Only @PirateJC and @RaananW should merge.

May contain errors in links and redirects.

Workflow section folders and files for organization possibilities only, content for this area is not my strong point.

PR #560, opened 2022-05-14
Create a top-level Contributing page

rdw-software:
IMHO, this is critical information that will be of interest to anyone who might consider becoming an active contributor, especially new users who haven't contributed before. Hiding it deep down in the hierarchy of documentation pages seems unwise. Removing barriers to entry like this should increase the amount and quality of contributions that Babylon can attract.

---

I only linked to the existing documentation pages as I didn't want to shuffle things around too much without your input, but I think it would be an improvement if those pages were to be moved to this top-level and highly-visible category, split into multiple sub-pages for the different tasks, and organized in a way that's less daunting and chaotic to help new contributors out.

I've seen #483 , but it's unclear what the status is and there are way too many changes to reasonably digest. It does seem like it's doing just what I suggested, albeit slightly differently. Maybe we can combine the ideas and make smaller changes iteratively?

### DIFF
--- a/configuration/structure.json
+++ b/configuration/structure.json
@@ -1251,6 +1251,13 @@
             "friendlyName": "Playground search",
             "children": {},
             "content": ""
+        },
+        "contribute": {
+            "friendlyName": "Contribute to Babylon.js",
+            "children": {
+
+            },
+            "content":"landing_pages/contribute_to_babylonjs"
         }
     },
     "navOverrides": { "previous": "" }
--- a/content/landing_pages/contribute_to_babylonjs.md
+++ b/content/landing_pages/contribute_to_babylonjs.md
@@ -1,0 +1,21 @@
+---
+title: Contribute to Babylon.js
+image:
+description: Learn how you can get involved with Babylon.js.
+keywords: babylon.js, contribution
+further-reading:
+video-overview:
+video-content:
+---
+
+# Contribute to Babylon.js
+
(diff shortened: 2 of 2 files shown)


RaananW:
Any issue with this page - https://doc.babylonjs.com/divingDeeper/developWithBjs/howToStart ?
I am not against adding the information, but it can be added to an already-existing page that is very much relevant to it :-)

rdw-software:
> Any issue with this page - https://doc.babylonjs.com/divingDeeper/developWithBjs/howToStart ? I am not against adding the information, but it can be added to an already-existing page that is very much relevant to it :-)

The idea was to have a top-level "entry" page and then lead people to the more detailed information. Ideally, the entire category of "contribution-related" info would be under this page, which is akin to what @BabylonJSGuide has done in the WIP PR (I believe). Maybe he could give his rationale for the change, as I suspect it's similar to my own line of thinking:

I tried to find information about how to contribute, and I eventually found it buried deep within a nested hierarchy of (unrelated) topics. I simply figured, it should be much more accessible since contributing is a key topic for any open source project :)

If you're asking, any issues with the contribution page itself, then yeah - several in fact. For example, it's enormous (=intimidating for new contributors), and it covers multiple different categories of tasks all at once, such as "setting up", "build process", "repository structure", "testing", "PR process", and likely more. I would split those and add subpages for each topic, as they're useful in different situations and to different people. However, that likely warrants further discussion, and I didn't want to make any sweeping changes.

BabylonJSGuide:
I think this is a good idea. In fact the possible reorganization of the docs has put contibuting at the top level.

Part of my rationale was to clearer split sections of the docs into related areas.

PirateJC:
Closing this PR. As @BabylonJSGuide highlighted, this exact issue will be addressed in the following PR: 

https://github.com/BabylonJS/Documentation/pull/622

PR #561, opened 2022-05-14
Increase the contrast of the Table of Contents component

rdw-software:
This makes the TOC stand out visually and the text easier to read. The padding is derived from the padding-bottom that was already in place, though I couldn't find a way to access that value directly (which would be preferable in case of future changes).

The heading uses a 13px margin so that's a bit inconsistent, but I'm not sure if it makes much of a difference visually and so I've left it as-is.

---

Before:

image

After:

image

As with any visual change, it's at least partially based on personal preference (though the goal was to create a visual separation between a page's table of contents and the actual contents themselves). Feel free to disregard if you don't think this is an improvement.

### DIFF
--- a/components/contentComponents/tableOfContent.component.tsx
+++ b/components/contentComponents/tableOfContent.component.tsx
@@ -10,6 +10,7 @@ const styles = makeStyles((theme: Theme) =>
             border: "1px solid #dedede",
             borderRadius: "5px",
             backgroundColor: "#f7f7f7",
+            marginBottom: "1.5rem",
             paddingLeft: "16px",
             paddingRight: "16px",
             zIndex: 100,
--- a/components/contentComponents/tableOfContent.component.tsx
+++ b/components/contentComponents/tableOfContent.component.tsx
@@ -7,6 +7,11 @@ const styles = makeStyles((theme: Theme) =>
     createStyles({
         contentRoot: {
             background: "#fafafa",
+            border: "1px solid #dedede",
+            borderRadius: "5px",
+            backgroundColor: "#f7f7f7",
+            paddingLeft: "16px",
+            paddingRight: "16px",
             zIndex: 100,
             width: "unset !important",
             "& h2": {


RaananW (approved):
LGTM. I'll wait for @PatrickRyanMS's opinion on this one.

rdw-software:
I noticed an issue that's already present in the documentation, but became very visible with this change:

Some pages have a leading paragraph before the first markdown heading. If one is present, the TOC component appears too close to the paragraph:

image

I can add a margin to the bottom of the TOC to fix this:

image

Unfortunately, with the extra margin pages that don't have a leading paragraph before the first heading look off:

image

The reason for this conundrum is that ``p`` elements have a similar ``margin-bottom`` as headings have ``margin-top``.

FWIW, I think having the extra ``margin-bottom`` looks better overall. It's certainly better than having too little space. Ideally, all pages would have a leading paragraph (or none), but since there's almost 700 pages fixing them all at once seems like a bit much...

For now, I'll add the extra margin so that you can compare both versions.

---

To fix the inconsistency itself, a template could be created for the structure of each page and then all pages that don't yet follow it slowly adapted to it, perhaps as part of the reorganization (#483 ). Either way, it's out of scope for this PR, so I'll leave it as-is.

PatrickRyanMS:
If the goal is to visually separate the ToC from the rest of the content, this would not be the option I would pursue. It departs with the original design in several ways:

- We are not using corner radiuses anywhere
- We are not using thin strokes around any content
- We are not using any low contrast elements due to accessibility concerns
- We are not using styling elements on many parts of the content to make the ones that have it (dark boxes around code or the purple boxes on links) stand out more and the user can expect what each one means.

image

Ideally, there should not be text after the ToC without a new  header to separate the text from the ToC to create a visual separation to set the ToC apart. But if the goal is to really make the ToC stand apart from the rest of the text and have more emphasis than the rest of the page, there are different ways to accomplish that with a higher contrast background that fills the <div> containing the ToC from side to side. However, that increased emphasis will always be there drawing attention when the user is reading any content with the ToC visible, so there would have to be a balance of not becoming the most important element on the page.

RaananW:
Closing this for now, please re-open or reply to the last comment, which is the important part here :-)

rdw-software:
Whoops! My bad, I thought I had already replied. Looks like I forgot.. sorry about that.

Thank you @PatrickRyanMS for the detailed review! I thought about my reason for coming up with this idea in the first place, and I believe that changing the layout isn't really needed. The actual problem is that there's inconsistent spacing, i.e., this here:

> Ideally, there should not be text after the ToC without a new header to separate the text from the ToC to create a visual separation to set the ToC apart.

I'm inclined to think that detecting any text being present before there's a sub-heading might be a better way to tackle the problem. I wanted to write a script to automatically check the markdown files, so that some consistency could be enforced (or at least a warning of some sort displayed to authors). Do you think that's a sensible approach? Maybe it's too strict, or not useful.

I suppose it would be easier (for me as an artistically-challenged person) than modifying the layout, with no unintended side effects.

PatrickRyanMS:
I'm going to ping @PirateJC about the question about the need for a script to detect text after the TOC as he is far more familiar with the content of the doc site. I don't know if this is something widespread or if there are only a few places where this happens.

PR #566, opened 2022-05-19
docs: remove `markAllAsDirty` in Material_plugins.md

yuqianma:
related discuss: https://forum.babylonjs.com/t/doc-markallasdirty-is-unnecessary-in-material-plugins/29494


Popov72:
I think you should also update the example playgrounds to remove the reference and use `markAllDefinesAsDirty` instead.

yuqianma:
> I think you should also update the example playgrounds to remove the reference and use `markAllDefinesAsDirty` instead.

Oh, right !

yuqianma:
@Popov72  Done?

RaananW (approved):
LGTM, waiting on @Popov72

Popov72:
That's ok for the PGs but you should also rename the pictures.

For eg, for `#GC63G5#3`, you should rename `public/img/playgroundsAndNMEs/pg-GC63G5-3.png` into `public/img/playgroundsAndNMEs/pg-GC63G5-16.png`.

deltakosh (approved):

yuqianma:
@Popov72 
Images are renamed.

I thought images are generated automatically by CI. 😂 So what if these images have to be replaced? Run build?
Maybe there should be some commit guides.
At the begining I modified the code within github directly. Later I opened a codespace. It would be sweet to make the process easily. As some are just typos or tiny updates like this one.

Popov72:
I think the picture is generated automatically locally, not on the build server: you would still have to commit and push it.

In the present case, as the PGs themselves have not changed, it's easier to simply rename the files.

RaananW:
> @Popov72 Images are renamed.
> 
> I thought images are generated automatically by CI. 😂 So what if these images have to be replaced? Run build? Maybe there should be some commit guides. At the begining I modified the code within github directly. Later I opened a codespace. It would be sweet to make the process easily. As some are just typos or tiny updates like this one.

`npm run build:local` will generate the images.

yuqianma:
Ready to merge?

PR #584, opened 2022-06-11
fix suffixes

aleksandr-vin:


RaananW (approved):
great catch! thank you so much :-)

PR #585, opened 2022-06-11
Fix external link to wikipedia

aleksandr-vin:


RaananW (approved):
you're awesome, thanks!

PR #593, opened 2022-06-27
change spaced parameter names to snake-cased

pr-o:
Hi, I've noticed some function parameter names have spaces within themselves.

I understand it is for the sake of readability, but IMHO, it might cause confusion to some readers.

I, for one, had to re-read the function signature while careful reading.

Thus, I propose the following changes.

- **AS IS**

- **TO BE**

&nbsp;

I personally think these changes are better as they explicitly show parameter boundaries, especially when a certain parameter gets line-wrapped as in the following example (see `target position`):

### DIFF
--- a/content/How_To/animation/animation_method.md
+++ b/content/How_To/animation/animation_method.md
@@ -12,7 +12,7 @@ video-content:
 
 ## Creating the animation
 ```javascript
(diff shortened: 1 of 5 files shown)


RaananW (approved):
LGTM!!!

PR #598, opened 2022-07-01
Update WebXR_Session_Manager.md

yuripourre:
Fix method name: https://github.com/BabylonJS/Babylon.js/blob/master/packages/dev/core/src/XR/webXRSessionManager.ts#L318


RaananW (approved):
Thanks!

PR #643, opened 2022-09-13
Fixing playground for chapter7/adding shadows

pavelfryz:
moved definition of ShadowGenerator inside ImportMesh callback.


sebavan:
Ohhhh I do not think we should fix the doc here, but the code of babylon :-) I ll work on it ASAP.

sebavan:
A fix is in progress, I will close the PR for now and all should be back on track later tonight :-)

Thanks a lot for letting us know !!!

PR #655, opened 2022-09-22
Contributing article: Using Vite with Babylon.js (updated)

wavetro:
Here is a re-done branch with my write-up of using Vite with Bablyon.js! @RaananW I hope this is implemented correctly now


RaananW:
This PR has a very interestint issue - a file that you added as an example (I assume to download?) is being parsed by the build system.
Would it be possible to add the file in a zip archive? Or - if you want it to be hosted and "loadable", you will need to provide a javascript file and not a typescript file. the typescript content can be in the doc pages, and the compiled .js file could be in the public directory

wavetro:
I think I'll relocate all those example files (html/ts/json) directly into the lesson's markdown as codeblocks instead. Thanks for the feedback!

wavetro:
Okay to merge? @RaananW @PirateJC

RaananW (approved):

PR #660, opened 2022-09-27
Add Section on loading draco files locally

stefanpl:
In a privacy-aware environment, loading assets from 3rd party servers is a no-no. I just spent some time trying to figure out why and where babylon was making requests against `preview.babylonjs.com`, and how to prevent it from happening.

As far as I understand, this only happens when using the gltf loader – so I added it here.

This is a mere proposal. Feel free to add/change/move/whatever-you-see-fit.

✌️


RaananW (approved):
Thanks 😊

PR #663, opened 2022-09-29
Misc Doc Fixes/Updates

PirateJC:


content/features/featuresDeepDive/mesh/billboard.md
```diff
@@ -0,0 +1,30 @@
...
+
+Billboard Mode Values:
+
+value|Type
+-----|-------------
+0 | BILLBOARDMODE_NONE
+1 | BILLBOARDMODE_X
+2 | BILLBOARDMODE_Y
+3 | BILLBOARDMODE_Z
+4 | BILLBOARDMODE_ALL
```
Popov72, line 28: The values are 0/1/2/4/7. There's also `BILLBOARDMODE_USE_POSITION = 128` (Object will rotate to face the camera's position instead of orientation).

I think we should tell that the values are a bit field and can be combined with a **OR** (**||** operator).

PR #701, opened 2022-11-07
Bones and skeletons: Update performance section

Popov72:
See https://forum.babylonjs.com/t/baked-animation-performance/32023/11


carolhmj (approved):
Thanks a ton ❤️

PR #716, opened 2022-11-28
Group VS Code topics together

rjamesak:
VS Code information has been moved so that the sections can be found together in the documentation.


sebavan (approved):

sebavan:
@bghgary should be able to have a look shortly

PR #744, opened 2022-12-27
Add cuboid with soft edges example to Ribbons Page

teamhitori:
Added extra whitespace to end of line so that lines breaks appear correctly in html output
Added cuboid with soft edges example to Ribbons Page


Popov72:
You should remove the lock files from the PR.

teamhitori:
Cool, my bad. I've removed those files now. Cheers -

sebavan (approved):

PR #750, opened 2023-01-17
fix: typos

PhilippeMorier:


RaananW (approved):
Thanks a lot for that. feels like it happened because of an incorrect find-and-replace caused it at a certain point.

PR #763, opened 2023-01-28
Fix typos in the "Prerequisites for Contributing" page

phildremi:
Separate PR since it requires renaming the entire page 😱


configuration/structure.json
```diff
@@ -2501,10 +2501,10 @@
                     "friendlyName": "Contribute to Babylon.js",
                     "content": "contribute/toBabylon",
                     "children": {
-                        "Prerequisits": {
-                            "friendlyName": "Prerequisits for Contributing",
+                        "Prerequisites": {
```
Popov72, line 2504: I think we can't make this change, else it will break existing links pointing to this page.

cc @RaananW

phildremi: It seems that there are no other references to this page inside the documentation:

image

Nor in the forums (if that's what you were thinking about):

image

Even Google just has this one result:

image

It does seem there's a `sitemap.xml` that I missed (will fix in a separate commit), but nothing else on GitHub either:

image

Popov72: I was not thinking of references inside the documentation but references that could have been used/shared by users, either as bookmarks or inside some docs for eg (or sent by emails, ...).

phildremi: Gotcha. Though I doubt anyone would bookmark [this minimalistic page](https://doc.babylonjs.com/contribute/toBabylon/Prerequisits), it's of course your decision :)

Perhaps it's possible to set up an alias somehow so that bookmark visitors will be redirected?

RaananW: There is a redirection json file that you can use to redirect from the old URL to the new one. Would be great if you add it there!
it will require an approval from an admin (because it is a configuration json), but that shouldn't be a problem.

public/sitemap.xml
```diff
@@ -107,7 +107,7 @@
 <url><loc>https://doc.babylonjs.com/contribute/contributeToDocs/easydoc</loc><lastmod>2022-12-13T12:45:57.562Z</lastmod></url>
 <url><loc>https://doc.babylonjs.com/contribute/contributeToDocs</loc><lastmod>2022-10-17T11:20:30.293Z</lastmod></url>
 <url><loc>https://doc.babylonjs.com/contribute/toBabylon/HowToContribute</loc><lastmod>2022-12-13T12:45:57.563Z</lastmod></url>
-<url><loc>https://doc.babylonjs.com/contribute/toBabylon/Prerequisits</loc><lastmod>2022-09-22T12:30:30.282Z</lastmod></url>
+<url><loc>https://doc.babylonjs.com/contribute/toBabylon/Prerequisites</loc><lastmod>2022-09-22T12:30:30.282Z</lastmod></url>
```
RaananW: sitemap is updated on a weekly base in an external process. You can revert this change

RaananW (commented):
Looks great! A redirect is missing, sitemap needs to be reverted.

Thanks a lot!

public/sitemap.xml
phildremi: Done. BTW: Is the sitemap update process documented anywhere? I can't seem to find any info about it in the docs.

I'm just curious, but I guess it might be worth documenting if it's something that contributors would need to know about.

configuration/structure.json
phildremi: I think I found the right file, but I guess the redirects don't work locally. so a deployment on Vercel is the only way to test?

RaananW: yes, these are server-side redirects. LGTM! thanks a lot

RaananW (approved):

public/sitemap.xml
RaananW: it's not documented intentionally :-)

But running locally `npm run dev:local` will generate the sitemap

PR #765, opened 2023-01-28
Clarify the build process for .fx shader assets

phildremi:
I meant to just fix a typo here (lowercase a), but I suppose the phrasing left me scratching my head:

If shaders are just assets that are being processed (subject) by the build tool (actor), then they can't generate (action) but rather are generated (passive)... Or did I misunderstand the process?

And yes, I may or may not be overthinking the language at times.


Popov72 (approved):

RaananW (approved):
The wording is better, yet :-)
When consumed, a .fx file will be transpiled to a typescript file that can be consumed in a typescript project. That change is correct, thanks a lot :-)

PR #776, opened 2023-02-09
Additional help added to Vite Basics tutorial

wavetro:
Added some extra instructions for vanilla JS users for my [Vite](https://doc.babylonjs.com/guidedLearning/usingVite) guide on the docs!

### DIFF
--- a/content/guidedLearning/usingVite.md
+++ b/content/guidedLearning/usingVite.md
@@ -48,7 +48,7 @@ Once you have your code editor and Node.js installed, we can begin setting up yo
 
 This installs Vite and some of Babylon.js' features in your created folder (in a subdirectory called `node_modules`.)
 
-**(5)** Now enter the command `npm init vite`. Enter the name of your Babylon.js project, such as `testProj` or whatever name you want. You can also set a package name, but the default one works fine. Choose "Vanilla" from the list that pops up, and then choose "TypeScript." (This setup also works with JavaScript, but with slightly different steps not documented here.)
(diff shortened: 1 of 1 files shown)


RaananW:
I am not sure how much these changes are needed? The code is technically the same, but without types. The other changes are cosmetic.
A short sentence stating that JavaScript users will need to remove the types would be enough, no? Or am I missing something?

wavetro:
@RaananW The original tutorial was written for very new programmers who are only familiar with coding the Start Your Journey chapters in the Playground, and don't know how to move that knowledge into an IDE or code editor. I then had someone approach me who's been learning JavaScript for the first time by learning Babylon (who wanted a non-TS version of this documentation,) so putting an entire JS version of the code here also doubles as a way to show those readers that TS is not very different from JS.

Let me know if you still want me to condense this into simpler instructions though!

RaananW:
This will be a very confusing tutorial for someone who is just getting starting with JavaScript. It's not that i mind having it, i just think the way it's presented it's not helping as it should. If you are providing a JavaScript tutorial, do it fully. Otherwise newcomers will be lost. "Do this step, but if you are using JavaScript do what's in the parenthesis, except for step 9, which you  should copy from down below".

Just my two cents, to be honest. If you say that you think this is the right way to go I'll approve it and merge it.

wavetro:
Fair point, would it be better if I just link to the forum post with all the differing instructions for JavaScript users? It would be cleaner than trying to implement them directly into the same page as the TypeScript instructions like I tried to do here

PR #779, opened 2023-02-16
Update material.md

yinzc:
[delete][what]: delete new


Popov72 (approved):
Thanks for catching it!

PR #784, opened 2023-02-25
Include section about large skyboxes

stefanpl:
If I understand them correctly, I think both options (maxZ and infiniteDistance) have their uses and should be mentioned here.

I've been working with a skyboxed scene for some time, and finally took the time to investigate why it didn't quite behave as I wanted. [This old SO post](https://stackoverflow.com/questions/35387282/how-to-make-a-very-large-skybox-babylon-js) was all the info I needed.


RaananW (commented):
Whereas I understand the need for that, it should just be a comment in the code that creates the skybox. It has less to do with the infiniteDistance variable and more to do with the user's scene configuration. A skybox that is larger than 1000 is probably needed, if your scene size is larger than 1000 units.

You can add a sentence saying "when creating the skybox make sure that your camera's maxZ is adjusted to the size of the skybox"

sebavan:
@stefanpl can you add the requested change ? I will merge right after.

sebavan:
Closing due to inactivity, let s reopen if updated ?

PR #808, opened 2023-04-03
Physics V2 

carolhmj:
- TODO:
- [x] Add a top-section page for "Legacy"
- [x] Add a "Your first physics scene" section in the "Using engines" page
- [x] Add a "Migrate from V1 Physics" page
- [x] On the Thin Instances section, document that setting a transformation at the mesh level is NOT supported


content/features/featuresDeepDive/physics.md
```diff
@@ -20,8 +48,8 @@ There are plugins for 3 physics engines:
```
Popov72, line 49: post => port

content/features/featuresDeepDive/physics.md
```diff
@@ -12,6 +12,34 @@ video-content:
...
+## Physics V2
+
+There are plugins for X physics engines:
+
+1.
+
+All need to be enabled before use.
+
+This architecture allows finer grain control on your physics scene.
+By making the difference between Collision shapes and Body dynamic.
```
Popov72: I think it should be the same sentence?

```
This architecture allows finer grain control on your physics scene by making the difference between Collision shapes and Body dynamic.
```

content/features/featuresDeepDive/physics.md
```diff
@@ -12,6 +12,34 @@ video-content:
...
+A typical use case is to have only one collision shape used by numerous bodies.
+This allows a lower memory footprint and more efficient memory addressing.
+A special effort has been put into better performances for a huge number of bodies. This is the case for the collision callback mechanism.
+
```
Popov72, line 39: I'm not sure to understand the last sentence wrt to the previous ones. Do you mean:

```
This is also the case for the collision callback mechanism.
```
?

content/features/featuresDeepDive/physics/v2/shapes.md
```diff
@@ -0,0 +1,82 @@
...
+```javascript
+const shape = new BABYLON.PhysicsShapeConvexHull(
+    mesh,   // mesh from which to produce the convex hull
+    scene   // scene of the shape
+);
+```
+
+```javascript
+const shape = new BABYLON.PhysicsShapeMesh(
+    mesh,   // mesh from which to produce the convex hull
```
Popov72: Probably wrong comment?

content/features/featuresDeepDive/physics/v2/forces.md
```diff
@@ -0,0 +1,109 @@
...
+
+// .... OR .....
+
+body.applyImpulse(
+    new BABYLON.Vector3(0, 10, 0), // direction and magnitude of the applied impulse
+    new BABYLON.Vector3(0, 0, 0) // point in the body's LOCAL space where the impulse will be applied    
+);
+```
+
+A reminder that, if a body's node is an InstancedMesh, you can choose which instance to apply the force/impulse by passing `instanceIndex` as a parameter:
```
Popov72: Don't you mean a thin instance instead of `InstancedMesh` here?

Also, as a total noob about physics engine, I wonder:
* as the force is continuous in the case of `applyForce`, how to remove it when I no longer want it?
* can we call `applyForce` multiple times for the same body? If yes, is it possible to remove some forces selectively afterwards?

content/features/featuresDeepDive/physics/v2/forces.md
```diff
@@ -0,0 +1,109 @@
...
+```
+
+## Physics Helper
+
+### Radial explosion impulse/force & gravitational fields
+
+You have the ability to create radial explosions & gravitational forces.
+
+The forces are never applied to impostors that have mass equal 0 (the ground for example).
+
```
Popov72, line 51: Do you mean "bodies" instead of "impostors"?

Also, I would have put this sentence in the first paragraph of the page. Or even as the very first sentence, before the paragraph, as a warning/note, as that seems important to me.

content/features/featuresDeepDive/physics/v2/compounds.md
```diff
@@ -0,0 +1,40 @@
...
+
+```javascript
+const myMesh = BABYLON.SceneLoader.ImportMeshAsync(...);
+
+const sphere = BABYLON.MeshBuilder.CreateSphere("sphere", { diameter: 2, segments: 32 }, scene);
+sphere.position.y = 3;
+sphere.parent = myMesh;
+const box = BABYLON.MeshBuilder.CreateBox("box", { height: 2, width: 1, depth: 1 });
+box.parent = myMesh;
+
```
Popov72, line 29: I think it would be clearer to create two `TransformNode` and not two meshes.

At first I did not understand why we would pass a mesh to `PhysicsShapeContainer.addChild`. I had to look at the signature of the method to understand that it expects a `TransformNode`, not a `Mesh`.

content/features/featuresDeepDive/physics/v2/aggregates.md
```diff
@@ -0,0 +1,50 @@
...
+      url: /features/featuresDeepDive/physics/forces
+video-overview:
+video-content:
+---
+
+# Physics Aggregate
+
+## What is an Aggregate
+
+The Physics Aggregate is a compound object that contains a Body and a Shape. It's a helper that allows creating all the objects necessary to physicalize your scene in just one call.
```
Popov72: The "compound" word is confusing because it is already used to mean a set of shapes.

I think you should use "container" instead of "compound" when you speak of a set of shapes (also, it matches the name of the `PhysicsShapeContainer` class) and use compound only wrt physics aggregate.

content/features/featuresDeepDive/physics/v2/compounds.md
```diff
@@ -0,0 +1,40 @@
+---
+title: Physics compounds
```
Popov72, line 2: I think "compounds" should be replaced by "container" everywhere, see my other comment in the Aggregate page.

Alternatively, to avoid too many changes, replace the "compound" word in the Aggregate page by something else.

content/features/featuresDeepDive/physics/v2/aggregates.md
```diff
@@ -0,0 +1,50 @@
...
+In that sense, it's similar to Physics V1 Impostors, but with more control.
+Once created, you can easily get the individual parts and tweak them.
+
+## How to use it
+
+```javascript
+const sphere = BABYLON.Mesh.CreateSphere("sphere", 16, 2, scene);
+const aggregate = new BABYLON.PhysicsAggregate(sphere, BABYLON.ShapeType.SPHERE, { mass: 1 }, scene);
```
Popov72: I think `ShapeType` should be renamed `PhysicsShapeType`, as all other types related to physics have `Physics` as a prefix.

content/features/featuresDeepDive/physics/v2/migrateFromV1.md
```diff
@@ -0,0 +1,60 @@
...
+const sphere = BABYLON.MeshBuilder.CreateSphere("sphere", { diameter: 2, segments: 32 }, scene);
+const aggregate = new BABYLON.PhysicsAggregate(sphere, BABYLON.ShapeType.SPHERE, { mass: 1, friction: 0.2, restitution: 0.3 }, scene);
+```
+
+Note we switched the constructor from `PhysicsImpostor` to `PhysicsAggregate`, and the second parameter (the type of the physical shape) from `BABYLON.PhysicsImpostor.SphereImpostor` to `BABYLON.ShapeType.SPHERE`. [Here are the available types of shapes](/features/featuresDeepDive/physics/rigidBodies#shape).
+
+## Option 2: No Aggregates
+
+One of the most important new features in the new architecture is the ability to fine-tune the collision shapes and reuse them, greatly increasing speed and customizability. So it is also possible to separately create bodies and shapes:
+
```
Popov72, line 36: There are other places in the doc where it says that reusing the same collision shape multiple times increases the speed, but I don't understand how? Even if two meshes use the same collision shape, the calculation to determine the collisions will be the same as if the two meshes used two different shapes?

content/features/featuresDeepDive/physics/v2/perfTips.md
```diff
@@ -0,0 +1,35 @@
...
+video-overview:
+video-content:
+---
+
+# Performance Tips
+
+## Reduce, Reuse, Recycle
+
+The main point that should be taking advantage of when using the API is to reuse as much objects as possible! For example, if you have a bowling game, instead of disposing and recreating your ball and pins every shot, you should keep the same objects and just reset their positions to the "default" state. By default, when you change the node that's associated to a body, there will be no effect. That's because the pre-step of updating the physic body with the values from the transform node is disabled by default, to improve performance. You can solve this by enabling the pre-step for a frame, and then disabling it again next frame:
+
```
Popov72, line 16: `when you change the node`: I think `when you change the properties of the transform node` would be clearer.

Popov72 (commented):
Very good doc!

I'm a total noob with physics engines and my English isn't much better, so take my comments with a grain of salt!

I think having a paragraph about the concept of "sleep" somewhere would be nice, because it's important for performance that physical objects don't consume resources if they don't have to. Or is this automatic in v2?

content/features/featuresDeepDive/physics/v2/aggregates.md
CedricGuillemet: Makes sense. I'm doing a PR for that.

content/features/featuresDeepDive/physics/v2/migrateFromV1.md
CedricGuillemet: Yes, it mostly improves memory usage.

content/features/featuresDeepDive/physics/v2/perfTips.md
CedricGuillemet: Note: doing so may or may not allow body to get to sleep mode, depending on Physics Engine.

content/features/featuresDeepDive/physics/v2/raycast.md
```diff
@@ -0,0 +1,33 @@
...
+var end = new BABYLON.Vector3(1, -20, 2);
+physicsEngine.raycastToRef(start, end, raycastResult);
+if (raycastResult.hasHit) {
+    console.log("Collision at ", raycastResult.hitPointWorld);
+}
+```
+
+Important thing to note here is the `raycastResult` that can (and should!) be reused between calls. This is the major difference with Physics V1 raycasting.
+
+** Example and PG **
```
CedricGuillemet: Should we go further here and describe fields of `raycastResult` ?

content/features/featuresDeepDive/physics/v2/collisionEvents.md
```diff
@@ -0,0 +1,45 @@
...
+
+  // world callback
+  const observable = plugin.onCollisionObservable;
+  const observer = observable.add((collisionEvent) => {
+    // Check if the player is involved in the collision
+    if (collisionEvent.collider === player || collisionEvent.collidedAgainst === player) {
+      // Process collisions for the player
+      // ...
+    } 
+  })
```
CedricGuillemet: Same as for the raycast, shall we describe the fields of `collisionEvent` or just point to the API doc ?

content/features/featuresDeepDive/physics/v2/compounds.md
carolhmj: I was mostly using "compounds" because it's the word used for V1, but the V2 equivalent is Container so I agree it sounds a bit strange... @CedricGuillemet I wonder if it's best to just use Container in this case?

content/features/featuresDeepDive/physics/v2/forces.md
carolhmj: Hmmm @CedricGuillemet can you elaborate a bit more in the applyForce case? From what I understand, even through the force application is continuous over time, the force is applied only until the next frame, right?

content/features/featuresDeepDive/physics/v2/forces.md
carolhmj: Good point, I've rearranged the paragraphs a bit and added the part about the zero mass on the first section.

content/features/featuresDeepDive/physics/v2/raycast.md
carolhmj: Hmmm since these fields can change with time, I wonder if it wouldn't be better to point to the codedoc instead?

content/features/featuresDeepDive/physics/v2/collisionEvents.md
carolhmj: I'd say to point to the API doc

content/features/featuresDeepDive/physics/v2/migrateFromV1.md
carolhmj: Good point, I'll reword it to explicitly mention memory usage.

content/features/featuresDeepDive/physics/v2/compounds.md
carolhmj: I was mostly using "compounds" because it's the word used for V1, but the V2 equivalent is Container so I agree it sounds a bit strange... @CedricGuillemet I wonder if it's best to just use Container in this case?

Popov72 (approved):

PR #820, opened 2023-04-19
Add more info to rigid bodies and shapes

eoineoineoin:


carolhmj (approved):
Amazing! Thank you so much for the incredible explanations 😄

PR #847, opened 2023-05-03
BabylonJS should go into dependencies not devDependencies

gsanta:
As you are using babylonjs in your compiled boundle not as a utility tool, it should go to the normal dependencies.  (As eslint also warns about it, so it is disturbing in vscode)


RaananW (approved):
That depends on the context of what you are doing, TBH.
If you are building your own app it doesn't really matter where you put it. As part of your build process babylon will either be included or not, irrelevant of the place in your package.json.
If you are building a plugin for babylon it should be in dev dependencies (and peer dependencies with the version your plugin supports).

I am approving this, as it is the correct path for a general project using babylon, but will probably add a small note later regarding where to place babylon when building your project.

PR #875, opened 2023-05-27
Update HowToContribute.md

RolandCsibrei:
Windows access denied on port 1337 troubleshooting

### DIFF
--- a/content/contribute/toBabylon/HowToContribute.md
+++ b/content/contribute/toBabylon/HowToContribute.md
@@ -213,6 +213,14 @@ The files to edit are sceneJs.js and sceneTs.ts in the source folder of the Baby
 
 As described in the next section, the Babylon server also offers a playground-snippet debugging.
 
+#### Windows permission denied troubleshooting
+
+If you get an error like `Error: listen EACCES: permission denied 0.0.0.0:1337` run the following commands as a privileged user:
+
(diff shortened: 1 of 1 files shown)


sebavan:
Interesting, I never ran into this :-) Adding @RaananW as I am not sure reset winnat is smthg we should advise on the doc without detailing what could go wrong ?

RaananW:
I totally agree with @sebavan. This is a local issue that is very system-dependent. I have never experienced that (and AFAIK no one from the team as well. This is the first time I enouncter an issue with a system that blocks access to port 1337. Might be a local firewall issue?
Anyhow - we shouldn't suggest resetting winnat in our official docs. this is an issue we can resolve on the forum based on the user's comment.

RolandCsibrei:
Guys, I was struggling to get it working for 3 hours. I tried different node versions,  I even uninstalled my antivirus despite I was double checking port usages, I wrote a simple node app to test whether it can connect to port 1337, I got through all recomendations I found on Google. The port was not used. The antivirus was not blocking the port. This solution worked. I suggest to leave this solution in the offcials docs but add some information what can the user check before triying this the last if nothing worked. How about that?

sebavan:
I am not against adding it in a troubleshooting note but we need an explanation of why and what can break doing this so at least ppl understand the risk if any ?

PR #931, opened 2023-07-14
Update masterPBR.md

### LINKED ISSUE #883: Documentation for setting up Specular textures directly in PBRMaterial when using metallic-roughness
Some of the parameters used for loading a mesh that utilizes KHR_specular are named in a way that can be confusing when trying to set up a similar material by hand. Look into the API descriptions as well as create some content in the appropriate section about what this is and how to set up.

From https://forum.babylonjs.com/t/f0-factor-and-reflection-textures-how-to-control-them-from-stingray-maya-materials/13584/1

PatrickRyanMS:
Add section about controlling specular reflections on metallic-roughness PBRMaterial. Explain parameters stemming from KHR_material_specular and similarly named parameters already existing in PBRMaterial and how they interact.

closes #883


content/features/featuresDeepDive/materials/using/masterPBR.md
```diff
@@ -96,6 +96,49 @@ Once the conversion done, let's see the custom options available on this version
...
 
+## Controlling Specular Reflections on Metallic-Roughness Materials
+There are times when having control over the power or color for specular reflections of a PBR metallic-roughness material is desirable. Further, there are substances that can’t be accurately rendered without this type of control. Typically, a PBR specular-glossiness material is used for these types of substances, but there are still ways to use a metallic-roughness material to render a substance which needs modification to its specular reflections. 
+
+For glTF models, there is a Khronos extension called `KHR_materials_specular` which adds `specular` and `specularColor` parameters to the metallic-roughness material. Additionally, the extension allows using textures to control these values through `specularTexture` and `specularColorTexture`. These parameters affect the dielectric BRDF to change the specular power or color of F0 reflections, previously only attainable using specular-glossiness materials. If a glTF file is authored with this extension, the Babylon.js loaders will translate the glTF material description to a `PBRMaterial` with the parameters set correctly.
```
bghgary: > `KHR_materials_specular`

A link to the spec would be good here?

content/features/featuresDeepDive/materials/using/masterPBR.md
```diff
@@ -96,6 +96,49 @@ Once the conversion done, let's see the custom options available on this version
...
+- **useOnlyMetallicFromMetallicReflectanceTexture**: If this parameter is set to true, only the Alpha channel of `metallicReflectanceTexture` is used in lighting calculations. This parameter is important if using textures authored for the `KHR_material_specular` extension because if both `metallicReflectanceTexture` and `reflectanceTexture` are assigned textures, `metallicReflectanceTexture` will be used by default unless `useOnlyMetallicFromMetallicReflectanceTexture` is set to true. If the parameter is set to true, the RGB channels of the texture assigned to `reflectanceTexture` will be used to determine values for `metallicReflectanceColor`.
+- **reflectanceTexture**: This is the same as `specularColorTexture` from `KHR_materials_specular`. The color stored in the texture’s RGB channels is multiplied with `metallicReflectanceColor` and scaled by `metallicF0Factor` to produce the final specular contribution **only if** there is no texture assigned to `metallicReflectanceTexture` **or** `useOnlyMetallicFromMetallicReflectanceTexture` is set to true. Otherwise, any texture assigned to `metallicReflectanceTexture` will have its RGB channels used in place of this texture.
```
bghgary: nit: I would recommend using italics for emphasis here instead of bold since the first word is bolded already.

```suggestion
- **reflectanceTexture**: This is the same as `specularColorTexture` from `KHR_materials_specular`. The color stored in the texture’s RGB channels is multiplied with `metallicReflectanceColor` and scaled by `metallicF0Factor` to produce the final specular contribution *only if* there is no texture assigned to `metallicReflectanceTexture` *or* `useOnlyMetallicFromMetallicReflectanceTexture` is set to true. Otherwise, any texture assigned to `metallicReflectanceTexture` will have its RGB channels used in place of this texture.
```

content/features/featuresDeepDive/materials/using/masterPBR.md
```diff
@@ -96,6 +96,49 @@ Once the conversion done, let's see the custom options available on this version
...
+
+To help illustrate how these parameters work together consider the following snippets of code which are all valid approaches to controlling specular reflections in a metallic-roughness material.
```
bghgary, line 114: I feel like we can have one or more playgrounds for this part? Is that easy to do?

bghgary (commented):
Looks good, though @sebavan should probably review the specifics.

content/features/featuresDeepDive/materials/using/masterPBR.md
```diff
@@ -96,6 +96,49 @@ Once the conversion done, let's see the custom options available on this version
...
+
+Babylon.js materials constructed as metallic-roughness can also exert control over specular highlights through several parameters on `PBRMaterial`. Beyond the basic setup of the material, these parameters need to be defined to shape specular reflections.
+
+- **metallicF0Factor**: This is the same as `specularFactor` from `KHR_materials_specular`. This parameter is multiplied with `metallicReflectanceColor` to control the intensity of specular reflections.
+- **metallicReflectanceColor**: This is the same as `specularColorFactor` from `KHR_materials_specular`. This parameter sets specular color at F0. As surface normals move toward F90, this value is interpolated to white.
```
Popov72, line 109: For maximum clarity, do we need to explain what F0 and F90 are, or do we expect the user to know what they mean?

Also, `metallicReflectanceColor` is the color at F90, not F0.

`As surface normals move toward F90, this value is interpolated to white` => it is interpolated to white by default (because 1 is the default value for `metallicF0Factor`), but in the general case it is interpolated to (`metallicF0Factor`, `metallicF0Factor`, `metallicF0Factor`).

content/features/featuresDeepDive/materials/using/masterPBR.md
```diff
@@ -96,6 +96,49 @@ Once the conversion done, let's see the custom options available on this version
...
+- **useOnlyMetallicFromMetallicReflectanceTexture**: If this parameter is set to true, only the Alpha channel of `metallicReflectanceTexture` is used in lighting calculations. This parameter is important if using textures authored for the `KHR_material_specular` extension because if both `metallicReflectanceTexture` and `reflectanceTexture` are assigned textures, `metallicReflectanceTexture` will be used by default unless `useOnlyMetallicFromMetallicReflectanceTexture` is set to true. If the parameter is set to true, the RGB channels of the texture assigned to `reflectanceTexture` will be used to determine values for `metallicReflectanceColor`.
```
Popov72, line 112: I'd put this last, as it refers to `reflectanceTexture`, which hasn't yet been explained.

content/features/featuresDeepDive/materials/using/masterPBR.md
PatrickRyanMS: I went with bold originally because italics between two monospaced words didn't stand out very well, but it's technically more accurate to go with italics here. Changing.

content/features/featuresDeepDive/materials/using/masterPBR.md
PatrickRyanMS: Yes, we should probably define F0 and F90. I didn't take the time because all other documentation around specular on our site and Khronos's repo assume the reader knows what they are. But we should take some time to explain. I will add that. 

For metallicReflectanceColor, I went back and forth on this. We say in the API it's color at F90, but the definition in the [specular extension](https://github.com/KhronosGroup/glTF/tree/main/extensions/2.0/Khronos/KHR_materials_specular#extending-materials) is "The F0 color of the specular reflection (linear RGB)." We say that the equivalent in Babylon to specularColorFactor is metallicReflectanceColor but our definitions are opposite one another. So I went to a test in the sandbox and if I set metallicReflectacnceColor to red, only the F0 specular color shows red and the F90 values are white. I am trying to strike a balance between the definitions and what people see. 

My struggle here is what you are saying makes complete sense when you look at the code, but not everyone (including me) will open the engine code and look so I am trying to find a balance between what the engine does and what the user perceives. Do you have a better way to resolve the opposite definitions between the Khronos extension and our docs?

content/features/featuresDeepDive/materials/using/masterPBR.md
PatrickRyanMS: done

content/features/featuresDeepDive/materials/using/masterPBR.md
PatrickRyanMS: done

content/features/featuresDeepDive/materials/using/masterPBR.md
PatrickRyanMS: I can make some PGs for them. It would require some external files for the textures, but I can make a couple quick examples.

content/features/featuresDeepDive/materials/using/masterPBR.md
Popov72: Actually, I think you're right and the doc for `metallicReflectanceColor` is wrong:
```doc
     * In metallic workflow, specifies an F90 color to help configuring the material F90.
     * By default the F90 is always 1;
     *
     * Please note that this factor is also used as a factor against the default reflectance at normal incidence.
     *
     * F0 = defaultF0 * metallicF0Factor * metallicReflectanceColor
     * F90 = metallicReflectanceColor;
```
According to the js/shader code, it seems it should be:
```doc
     * In metallic workflow, specifies an F0 color.
     * By default the F90 is always 1;
     *
     * Please note that this factor is also used as a factor against the default reflectance at normal incidence.
     *
     * F0 = defaultF0_from_IOR * metallicF0Factor * metallicReflectanceColor
     * F90 = metallicF0Factor;
```
cc @sebavan

content/features/featuresDeepDive/materials/using/masterPBR.md
PatrickRyanMS: I have made a PG showing three ways to set up specular as described in the section. Let me know what you think.

content/features/featuresDeepDive/materials/using/masterPBR.md
PatrickRyanMS: @Popov72, I added an opening paragraph explaining F0 and F90. Let me know what you think of that. @sebavan, what do you think about the comment above?

sebavan: yup good catch !!! let s change the code doc as well then

content/features/featuresDeepDive/materials/using/masterPBR.md
```diff
@@ -96,6 +96,53 @@ Once the conversion done, let's see the custom options available on this version
...
+
+There are times, however, when having control over the power or color of the specular reflections of a PBR metallic-roughness material is desirable. Further, there are substances that can’t be accurately rendered without this type of control. Typically, a PBR specular-glossiness material is used for these types of substances, but there are still ways to use a metallic-roughness material to render a substance which needs modification to its specular reflections.
```
bghgary: Sorry, noticed this on a second pass. I'm not sure this is typical anymore and probably isn't future proof even if it is right now. Maybe this?

```suggestion
There are times, however, when having control over the power or color of the specular reflections of a PBR metallic-roughness material is desirable. Further, there are substances that can’t be accurately rendered without this type of control. A PBR specular-glossiness material can be used for these types of substances, but there are ways to use a metallic-roughness material to render a substance which needs modification to its specular reflections.
```

content/features/featuresDeepDive/materials/using/masterPBR.md
bghgary: Looks good. Maybe labels might help, but not required.

bghgary (approved):

content/features/featuresDeepDive/materials/using/masterPBR.md
Popov72: I fixed the doc:

https://github.com/BabylonJS/Babylon.js/pull/14067

Popov72 (approved):

PR #933, opened 2023-07-24
Docs Fix: static call GetAvailableFeatures()

johannesvollmer:
Hi! I just quickly fixed this documentation in one minute. I think I fixed an issue: The documentation said `GetAvailableFeatures() ` was an instance method, but it's a static method, so we need to call it from the class. Sorry if this was already fixed, I admit I didn't check that.


RaananW (approved):

RaananW:
Thanks a lot!

PR #945, opened 2023-08-12
Document MeshDebugPluginMaterial

alexchuber:
Adding a documentation page for using the MeshDebugPluginMaterial. Let me know if this is a good place for it-- I wasn't sure where it should go in the structure.

(Yes, this is very last minute, but such is the beauty of GitHub and open-source. I'll still be available to make changes :))


configuration/structure.json
```diff
@@ -2513,6 +2513,11 @@
                             "friendlyName": "Display Axes",
                             "children": {},
                             "content": "toolsAndResources/utilities/World_Axes"
+                        },
+                        "VisualizeMeshData": {
```
Popov72: I would have put it under Features => Deep Dive => Mesh :)

Utility functions are intended for bits of external code to help do a specific thing. Here, it's more of a debugging option for meshes, that uses code already existing in the library.

Let's see what others think (cc @PirateJC).

Popov72 (commented):
LGTM!

configuration/structure.json
RaananW: I wonder if @alexchuber got the time to do that now. I think we might need to merge it and change it ourselves, but i'll wait a bit to see if she manages to answer this week

alexchuber: I'm here! Lol. I figured I was putting it in the wrong place, but I didn't have a good feel for the different sections in the docs.

alexchuber: (I'll wait a bit to see if cc'd agree with the new path before committing the change)

sebavan (approved):

PR #951, opened 2023-08-24
add missing option in highlightLayer docs

SalmaBesbes:
Hello community, 

I've always assumed that highlight layer cannot be displayed as a stroke, turned out I was missing that since it was not written in the docs.


sebavan (approved):
THANKS !!!!

PR #966, opened 2023-09-12
support NGE in the Docs

RaananW:
Just like NME, it is now possible to add NGE examples in a playground page using the following syntax:

```html
<NGE id="#MSH13Z" title="The default NGE" description="The default NGE saved as a snippet. Meaningless, but powerful!"/>
````

### DIFF
--- a/lib/frontendUtils/frontendTools.ts
+++ b/lib/frontendUtils/frontendTools.ts
@@ -5,7 +5,15 @@ export const getExampleLink = (example: Partial<IExampleLink>, full: boolean = t
     const id = idToUse ? (idToUse[0] === "#" ? idToUse : `#${idToUse}`) : "";
     // webgpu parameter. Stay safe and validate
     const params = (example.engine && example.engine==='webgpu') ? `?${example.engine}` : '';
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: lib/frontendUtils/frontendTools.ts [hunk 1]
```diff
@@ -5,7 +5,15 @@ export const getExampleLink = (example: Partial<IExampleLink>, full: boolean = t
     const id = idToUse ? (idToUse[0] === "#" ? idToUse : `#${idToUse}`) : "";
     // webgpu parameter. Stay safe and validate
     const params = (example.engine && example.engine==='webgpu') ? `?${example.engine}` : '';
-    return (example.type === "nme" ? "https://nme.babylonjs.com/" : "https://playground.babylonjs.com/" + (full ? "full.html" : "")) + params + id;
+    switch (example.type) {
+        case "nge":
+            return `https://nge.babylonjs.com/${id}`;
```
sebavan, line 10: Wouldn t we miss param which was in the nme address before ? Same for nme

RaananW: There are no parameters passed to nme. Just a snippet id. The only parameter supported in the playground is the webgpu flag

sebavan (approved):

PR #967, opened 2023-09-14
Allow anchoring an image with H2 and H3 elements

RaananW:
The idea is to allow anchoring a title with an image before the H tag.

This is now supported on H2 and H3 tags. Usage:

Instead of `##` or `###` use

```html
<H2Image title="The title Of this thing" image="/img/linkToImage.jpg"/>

OR

<H3Image title="The title Of this thing" image="/img/linkToImage.jpg"/>
```

### DIFF
--- a/components/markdownComponents/markdownComponents.tsx
+++ b/components/markdownComponents/markdownComponents.tsx
@@ -26,4 +26,6 @@ export const markdownComponents = {
     h2: H2MarkdownComponent,
     h3: H3MarkdownComponent,
     h4: H4MarkdownComponent,
+    H2Image: H2MarkdownComponent,
+    H3Image: H3MarkdownComponent,
 };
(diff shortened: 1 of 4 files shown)


Popov72 (approved):

PatrickRyanMS (commented):
I assume that along with title= and image=, we can also add an alt tag for the image as we would with any other HTML tag?

RaananW:
> I assume that along with title= and image=, we can also add an alt tag for the image as we would with any other HTML tag?

The image's alt tag is the title, but i can add an override of needed

PatrickRyanMS:
I think having the option of an alt tag override would be good. I can see cases where we may want the alt tag to be far more descriptive than what we want for the h2/h3 tag text.

RaananW:
> I think having the option of an alt tag override would be good. I can see cases where we may want the alt tag to be far more descriptive than what we want for the h2/h3 tag text.

It shall be done 😎

RaananW:
> I think having the option of an alt tag override would be good. I can see cases where we may want the alt tag to be far more descriptive than what we want for the h2/h3 tag text.

you can now add an optional `alt` parameter to override the alt for the image. if it is not provided, the title will be used.

Popov72 (approved):

PatrickRyanMS (approved):

PR #985, opened 2023-10-03
Expanding language about what can be animated

### LINKED ISSUE #984: Update language in Animation docs where targets are implied to be meshes
There are several places in the animation docs where we could expand the language around animation targets to be more agnostic of what is being animated. In many cases like tooltips, the target of the animation is specified to be a mesh which ignores the fact that animations can target almost anything which is why we need to specify the type of the animation (float, vector3, etc) since the animation knows nothing about the target. Making this language clear in the docs will help users see that animations are more flexible than a quick read of the docs implies.

PatrickRyanMS:
Our animation docs were very mesh-centric. Expanded language to include non-mesh objects that can be animated. I added @deltakosh to this review to ensure that the nuance of the language updated here is not inaccurate or misleading.

Closes #984


content/features/featuresDeepDive/animation/animation_design.md
```diff
@@ -20,20 +20,15 @@ The first stage design is to sketch what is needed at key time points, a little
...
 
-In Babylon.js the _Animation_ is changing the position of the box along the x axis and its x position is a floating point number and
-
-the _animation_ should loop. In code the animation which slides an item in the x direction becomes
+In Babylon.js the _Animation_ is changing the position of the box along the x axis and its x position is a floating point number and the _animation_ should loop. It is critical to declare the type for the animation because the animation has no concept of what the performer is. In this case, the animation is simply moving the box's position along the X axis. While the animation will know it's target - the box - it has no understanding of what it means to animate its position along the X axis. To combat this, we need to define the type for the animation as a floating point animation with `BABYLON.ANIMATION.ANIMATIONTYPE_FLOAT` and define the property which will accept the input as `position.x`. With these definitions, the animation has a roadmap of how to apply its keyframe data to the target resulting in a very flexible and powerful system that can animate almost anything. In code, this declaration looks like
```
carolhmj:
```suggestion
In Babylon.js the _Animation_ is changing the position of the box along the x axis, which is floating point number, and should also loop. It is critical to declare the type for the animation because the animation has no concept of what it is animating. In this case, the animation is simply moving the box's position along the X axis. While the animation will know its target - the box - it has no understanding of what it means to animate its position along the X axis. To combat this, we need to define the type for the animation as a floating point animation with `BABYLON.ANIMATION.ANIMATIONTYPE_FLOAT` and define the property which will accept the input as `position.x`. With these definitions, the animation has a roadmap of how to apply its keyframe data to the target resulting in a very flexible and powerful system that can animate almost anything. In code, this declaration looks like
```

content/features/featuresDeepDive/animation/animation_design.md
```diff
@@ -20,20 +20,15 @@ The first stage design is to sketch what is needed at key time points, a little
...
 const frameRate = 10;
 
 const xSlide = new BABYLON.Animation("xSlide", "position.x", frameRate, BABYLON.Animation.ANIMATIONTYPE_FLOAT, BABYLON.Animation.ANIMATIONLOOPMODE_CYCLE);
 ```
 
-The key frames are at 0, 1 and 2 seconds. To find the frame number after t seconds multiply the time by the frame rate, i.e. t x frameRate.  
-In this case the key frames are at frame numbers, 0, 1 x frameRate and 2 x frame Rate.
-
-Starting the box at x = 2 and sliding it to x = -2, gives the x positional values of the box after 0, 1 and 2 seconds as 2, -2 and 2 respectively.
+The key frames are at 0, 1 and 2 seconds. To find the frame number after t seconds multiply the time by the frame rate, i.e. t x frameRate. In this case the key frames are at frame numbers, 0, 1 x frameRate and 2 x frame Rate. Starting the box at x = 2 and sliding it to x = -2, gives the x positional values of the box after 0, 1 and 2 seconds as 2, -2 and 2 respectively.
```
carolhmj:
```suggestion
The key frames are at 0, 1 and 2 seconds. To find the frame number after t seconds multiply the time by the frame rate, i.e. `t x frameRate`. In this case the key frames are at frame numbers, 0, 1 x frameRate and 2 x frame Rate. Starting the box at x = 2 and sliding it to x = -2, gives the x positional values of the box after 0, 1 and 2 seconds as 2, -2 and 2 respectively.
```

carolhmj (commented):
Great updates! Just have a few nits suggestions on wording 😄

content/features/featuresDeepDive/animation/animation_method.md
```diff
@@ -37,26 +37,54 @@ const myAnim = new BABYLON.Animation(name, property, frames_per_second, property
...
+myAnimation.animations = [];
+myAnimation.animations.push(weightAnimation);
+
+```
+
+Remember, above we said that anything that the engine can reach can be animated with our animation system. This also applies to content outside of the canvas. For example, since we can reach any DOM object with Javascript, we can apply an animation to it. If we wanted to, we could use an animation to drive the border-radius of a `<div>` object.
+
+### beginAnimation
+
+Now that we have pushed an animation to a target object, starting the animation is a simple case of telling the scene to begin the animations stored with the target object.
```
Popov72: Maybe we should add a sentence to make clear that `beginAnimation` will actually start all animations stored in the `target.animations` array. It may not be obvious from the name of the method...

Popov72 (approved):
LGTM!

RaananW (approved):

RaananW:
Let me know if this can be merged. There are two open suggestions

content/features/featuresDeepDive/animation/animation_method.md
PatrickRyanMS: updated in https://github.com/BabylonJS/Documentation/pull/985/commits/a1bb49597e5396b9df7d5758de73276390ccbd5e

Popov72 (approved):

PatrickRyanMS:
> Let me know if this can be merged. There are two open suggestions

I merged the suggestions from @carolhmj and added additional context to cover the suggestion from @Popov72. If @Popov72 is good with the addition, we should be good to merge.

Issue #207, opened 2021-04-06
Add the ability to create notifications in docs


RaananW:
For example a warning that a feature is not available in XR or that a certain function is deprecated.
This will require a new markdown element that can be added to doc pages.

impurity-dev:
@RaananW What exactly did you have in mind for this? Would it be on the entire page?

RaananW:
The idea is to have some form of a markdown element that lets you add notes and remarks. Something similar to alerts in material design (https://mui.com/material-ui/react-alert/)

impurity-dev:
@RaananW can this be closed?

PR #1007, opened 2023-11-10
Add alert MDX custom component

impurity-dev:
image
The image shows the original warning, and the bottom shows the use of the @mui/material Alert component. I have added the custom <Alert ...> markdown to be used anywhere.

I did _not_ replace all instances of the alerts, as I wanted to ensure this custom component was good to go first. I did replace the glTF.md warning, as an example. If this merges, I plan to go through and update all alert instances as a follow-up.

Issue: #207

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -2,14 +2,9 @@
 # yarn lockfile v1
 
 
-"@aashutoshrathi/word-wrap@^1.2.3":
-  version "1.2.6"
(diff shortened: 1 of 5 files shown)


yarn.lock
RaananW, line 1: if no dependency was updated, it would be wonderful to revert this change

RaananW (approved):
This is GREAT!! thanks a lot for that. Just a small comment about the lock file

yarn.lock
impurity-dev: @RaananW No worries at all. Reverted 😄

PR #1017, opened 2023-11-21
Correct some sentence errors

wy-luke:


content/features/featuresDeepDive/particles/particle_system/particle_system_intro.md
```diff
@@ -10,7 +10,7 @@ video-content:
 
 ## An Introduction to the Particle System
 
-The creation of a particle system requires a name and its capacity and the number of particles in the system (capacity). The system is designed to produce particles that emit at a given rate, move and last for a set lifetime before they are re-cycled and re-emitted.
+The creation of a particle system requires a name and its capacity of the particles in the system. The system is designed to produce particles that emit at a given rate, move and last for a set lifetime before they are re-cycled and re-emitted.
```
sebavan: The sentence does not sound correct either.

```requires a name and its capacity of the particles in the system``` 

Should probably be smthg like 

```requires a name and its capacity. The capacity defines the max number of particles alive at the same time in the system```

wy-luke: Yes, you're right. My previous intention was to minimize changes as much as possible. 

I have corrected it in the latest commit. Thanks!

RaananW (approved):

PR #1018, opened 2023-11-22
Update steps in the “creation summary”,

fisicagames:
I noticed the lack of information about creating the ‘index.html’ file inside the ‘public’ folder. In the “creation summary”, I suggest adjusting the texts of steps 5. and 6. in Application Creation Summary.

### DIFF
--- a/content/setup/frameworkPackages/es6Support.md
+++ b/content/setup/frameworkPackages/es6Support.md
@@ -49,7 +49,7 @@ As the ES6 version is composed of many separated files the usual way to consume 
 2. Open **GitBash** (or similar) for **MyAwesomeApp**
 3. Run `npm init`
 4. Run `npm install webpack webpack-cli webpack-dev-server --save-dev`
-5. Run npm install @babylonjs/core and npm install --save-dev @babylonjs/materials
+5. Run `npm install @babylonjs/core @babylonjs/materials`
 6. Create Folder **public** in MyAwesomeApp and create **index.html** file in public as described in First App Section below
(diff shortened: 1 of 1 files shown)


content/setup/frameworkPackages/es6Support.md
```diff
@@ -49,8 +49,8 @@ As the ES6 version is composed of many separated files the usual way to consume
 2. Open **GitBash** (or similar) for **MyAwesomeApp**
 3. Run `npm init`
 4. Run `npm install webpack webpack-cli webpack-dev-server --save-dev`
-5. Run `npm install @babylonjs/core`
-6. Create **index.html** file in **MyAwesomeApp** as described in `First App Section` below
+5. Run npm install @babylonjs/core and npm install --save-dev @babylonjs/materials
```
RaananW: Removing the code section will make it confusing. This would be better:

```suggestion
5. Run `npm install @babylonjs/core @babylonjs/materials`
```

fisicagames: It's important to include the step that involves creating the index.html file inside the public folder. I've seen two topics on the forum for the same issue. Example: [https://forum.babylonjs.com/t/i-am-unable-to-open-babylon-locally-in-the-browser/42123 ](url).

RaananW: Not sure what you are referring to. I never said this shouldn't happen. all I said is that you removed the code definition. the ` characters. which will make it seem like part of the text, instead of copy-able command.

RaananW (approved):

Issue #790, opened 2023-03-09
Add "click on picture" feature to show it at full resolution


Popov72:
That would help when some details are too small to be seen on a small picture.

impurity-dev:
@RaananW Can we close this now?

Also what is the process for closing issues that have been addressed? I can look through and ping you, but don't wanna put your notifications on blast if that's too much.

PR #1020, opened 2023-11-27
Enhancement #790: Expand Image

impurity-dev:
## Issue
#790 

## Description
Added a hover effect to images in the doc that allows you to expand an image in a model. I'm sure the design could be refined at some point, but it gets the job done. It is enabled by default but if in the URL your used the queryparam `expandable=false` it will disable this feature.

⚠️ I did add a round border and a box shadow to the expandable images to give it a lil extra pizaz.

### Disabled
image

### Hoverable
image

### Hovering
image

### Expanded
image

### DIFF
--- a/components/markdownComponents/image.component.tsx
+++ b/components/markdownComponents/image.component.tsx
@@ -1,6 +1,6 @@
 import Image from "next/image";
 import { FunctionComponent, useEffect, useRef, useState } from "react";
-import { IconButton, Theme, Modal, Card, CardActions, CardContent, CardMedia } from "@mui/material";
+import { IconButton, Theme, Modal, Card, CardHeader } from "@mui/material";
 import { createStyles, makeStyles } from "@mui/styles";
 import { IImageEmbed } from "../../lib/content.interfaces";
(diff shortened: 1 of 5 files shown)


RaananW (changes requested):
Huzzah! thanks a lot :-)

The only change I ask for is to add a close button, because on very low resolutions there is no way to close the modal if the image is expanded. 
Also, it would probably be better to have a full-screen modal on mobiles to get the image as big as possible.

impurity-dev:
> Huzzah! thanks a lot :-)
> 
> The only change I ask for is to add a close button, because on very low resolutions there is no way to close the modal if the image is expanded. Also, it would probably be better to have a full-screen modal on mobiles to get the image as big as possible.

Thanks for the feedback! On it 👍🏼

RaananW:
> > Huzzah! thanks a lot :-)
> > The only change I ask for is to add a close button, because on very low resolutions there is no way to close the modal if the image is expanded. Also, it would probably be better to have a full-screen modal on mobiles to get the image as big as possible.
> 
> Thanks for the feedback! On it 👍🏼

Amazing work, as always!

RaananW (approved):

PR #1035, opened 2024-01-03
typesetting error in "The Animation Method" page

aleyx1979:
missing a CR/LF (or < br / >) before  BABYLON.Animation.ANIMATIONLOOPMODE_YOYO

image


aleyx1979:
I  think that i have fix this error and the md file is displayed correctly by github markdown viewer.
But in html page, it still exists.
Some magic happens on "converting markdown to html" ?

RaananW:
> I think that i have fix this error and the md file is displayed correctly by github markdown viewer. But in html page, it still exists. Some magic happens on "converting markdown to html" ?

This is happening by typedoc, possible they have markdown support that I missed. I will leave this open and assign me to see if there is something that can be done simply to fix that

RaananW:
Did you mean to open the PR, or just an issue?

PR #1046, opened 2024-01-25
Add simple lightmap example

carolhmj:


sebavan:
Wonder if @PatrickRyanMS has a softer lightmap texture ?

PatrickRyanMS:
@sebavan the only model I have that was set up to use lightmaps was https://forum.babylonjs.com/t/lightmaps-how-they-behave-in-sandbox-babylonjs/12357/3?u=patrickryan and it was using it as a shadow map only. No PG on this one. If you want to use this scene, I can quickly gather it together and make a PG, but I don't have anything else ready to go that uses lightmaps.

RaananW (approved):
Approved, but will let either @PatrickRyanMS or @sebavan merge it in case they have a better demo example

RaananW:
I am merging this. if there is a better demo we can always improve this

PR #1052, opened 2024-01-30
overhaul createscreenshot

### LINKED ISSUE #841: Update CreateScreenshot docs page
https://forum.babylonjs.com/t/documentation-for-screenshots-fails-to-relay-basic-info/40179

The page needs to explain better the difference between the CreateScreenshot methods.

carolhmj:
Closes https://github.com/BabylonJS/Documentation/issues/841
This attempts to explain better the difference between the two CreateScreenshot versions, and also add concrete examples. If there is more information that can be added please let me know! 😄

### DIFF
--- a/content/features/featuresDeepDive/scene/renderToPNG.md
+++ b/content/features/featuresDeepDive/scene/renderToPNG.md
@@ -8,220 +8,46 @@ video-overview:
 video-content:
 ---
 
-# How To Render a Scene to a PNG
+# How to
 
-First you may ask why is it better than "ctrl + prt scr" screenshot or the Snippet app?
-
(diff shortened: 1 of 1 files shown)


RaananW (commented):
you have removed a lot of existing text. is it because you thing is is not needed, or is it outdated?

carolhmj:
@RaananW I believe some parts are not necessary, like the creating gifs part, and I checked the contents of the hint and it does not seem to be true anymore: https://playground.babylonjs.com/#6UN291 (I'm taking a screenshot of an off-screen object and didn't have to set it as active mesh to work)

RaananW (approved):

PR #1054, opened 2024-02-08
First-Frame madness, solution suggestion

RaananW:
This should solve the first-frame issue with theme toggling.


sebavan (approved):

impurity-dev:
all hail @RaananW  👑

PR #1055, opened 2024-02-22
Update displayBoundingBoxes.md drawing bounding box documentation with getHierarchyBoundingVectors() example

dsaffo:
Added examples of using getHierarchyBoundingVectors() to draw bounding box around a node and its children as discussed in this thread: https://forum.babylonjs.com/t/update-parent-bounding-box-on-child-added/48200


dsaffo:
Seem to have messed up the headers on that page some how

Popov72 (approved):

PR #1063, opened 2024-03-03
Cleanup some XR input docs

sevaseva:
I just cleaned up language a bit as I was reading it for the first time. Did only a part of the page in this one.


RaananW (approved):
Thank you so much! :-)

PR #1077, opened 2024-03-12
Link to left-handed system explanation

sevaseva:
I didn't immediately find a perfect page on Wikipedia (many mention computer graphics only in passing; orders of axes often different), and out of many articles online, I thought a google search link is one of the safest over time. If there's a URL that fits better, please do replace with that

### DIFF
--- a/content/features/featuresDeepDive/cameras/webVRCamera.md
+++ b/content/features/featuresDeepDive/cameras/webVRCamera.md
@@ -308,7 +308,7 @@ Just like the WebVR camera, the controllers export their (right-handed!) raw pos
 
 ## Notes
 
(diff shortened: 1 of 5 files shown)


RaananW (changes requested):
I know it's in the description, but I would rather not put a google search query in our docs page :-)

This is something every user can search for themselves. I totally get that it is a new term for newcomers. If we find a good article, describing the two methods - let's add it. It's actually not that simple TBH - LHS and RHS can differ between two implementations, with the main difference being the negative/positive forward vector.

Also, since babylon is LHS per default, i guess it would also make sense to add a link to an RHS article :-)

sevaseva:
Fair enough. If I come across an appropriate page, i'll send another pr

PR #1120, opened 2024-04-29
feature: refactor get the snippet code

xxxigm:
Changes made:

1. Removed unnecessary nested promises by chaining then() calls.
2. Used forEach loop instead of for loop for iterating over codeLines.
3. Replaced indexOf with includes for checking if a line contains the search term.
4. Simplified the logic for setting startingLine and foundLine within the loop.
5. Replaced arrow functions with concise arrow functions where possible.

### DIFF
--- a/components/contentComponents/playgroundSearchResult.tsx
+++ b/components/contentComponents/playgroundSearchResult.tsx
@@ -103,32 +103,33 @@ export const PlaygroundSearchResult: FunctionComponent<{ searchResult: IPlaygrou
 
     useEffect(() => {
         // get the snippet code
-        fetch(`https://snippet.babylonjs.com/${searchResult.snippetIdentifier}/${searchResult.version}`).then((response) => {
-            response.text().then((text) => {
-                const parsed = JSON.parse(JSON.parse(text).jsonPayload);
-                const { code } = parsed;
-                const codeLines: Array<string> = (code || "").split("\n");
-                let startingLine = 0;
-                let foundLine = -1;
-                const lowerTerm = term.toLowerCase();
-                if (type === "code") {
-                    for (foundLine = 0; foundLine < codeLines.length; ++foundLine) {
-                        if (codeLines[foundLine].length > 100) {
-                            codeLines[foundLine] = `${codeLines[foundLine].substring(0, 100)}...`;
-                        }
-                        if (codeLines[foundLine].toLowerCase().indexOf(lowerTerm) !== -1) {
-                            startingLine = Math.max(foundLine - 5, 0);
-                            break;
-                        }
+        fetch(`https://snippet.babylonjs.com/${searchResult.snippetIdentifier}/${searchResult.version}`)
+        .then(response => response.text())
+        .then(text => {
+            const parsed = JSON.parse(JSON.parse(text).jsonPayload);
+            const { code } = parsed;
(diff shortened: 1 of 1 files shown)


components/contentComponents/playgroundSearchResult.tsx
```diff
@@ -103,32 +103,35 @@ export const PlaygroundSearchResult: FunctionComponent<{ searchResult: IPlaygrou
...
-                            codeLines[foundLine] = `${codeLines[foundLine].substring(0, 100)}...`;
-                        }
-                        if (codeLines[foundLine].toLowerCase().indexOf(lowerTerm) !== -1) {
-                            startingLine = Math.max(foundLine - 5, 0);
-                            break;
-                        }
+        fetch(`https://snippet.babylonjs.com/${searchResult.snippetIdentifier}/${searchResult.version}`)
+        .then(response => response.text())
+        .then(text => {
+            const parsed = JSON.parse(JSON.parse(text).jsonPayload);
```
yedpodtrzitko, line 109: would this work?

```suggestion
        .then(response => response.json())
        .then(jsonData => {
            const parsed = JSON.parse(jsonData.jsonPayload);
```

xxxigm: @yedpodtrzitko I think keeping my version will work fine. As for yours, I'm not sure.

components/contentComponents/playgroundSearchResult.tsx
```diff
@@ -103,32 +103,35 @@ export const PlaygroundSearchResult: FunctionComponent<{ searchResult: IPlaygrou
...
+            const lowerTerm = term.toLowerCase();
+    
+            if (type === "code") {
+                codeLines.forEach((line, index) => {
+                    if (line.length > 100) {
+                        codeLines[index] = `${line.substring(0, 100)}...`;
+                    }
+                    if (line.toLowerCase().includes(lowerTerm)) {
+                        foundLine = index;
+                        startingLine = Math.max(foundLine - 5, 0);
```
sebavan, line 121: Without the break, the code might not early exit anymore, why this change ?

xxxigm: @sebavan Wow, thank you for your question. Updated!

components/contentComponents/playgroundSearchResult.tsx
sebavan: We should try as it is better ? can you check @NguyenTuanCanh ?

components/contentComponents/playgroundSearchResult.tsx
sebavan: Your version is still different as foundLine will not be set outside of the condition but was always set in the loop version. I would in this case recommend to keep the loop as there is no reasons not too ?

components/contentComponents/playgroundSearchResult.tsx
RaananW: Both should work, one needs one less JSON.parse call.

RaananW:
This is mainly a cosmetic change, right? No functionality has changed?

RaananW:
I am a bit confused as to the current changes. Formatting has changed a bit, but it seems like the code as it is is almost (or even completely) the same?

RaananW:
any update here? or should I close it?

PR #1124, opened 2024-05-05
update mmd loader pg

noname0310:
Updated mmd loader's playground to match the latest package changes


content/communityExtensions/mmdLoader.md
```diff
@@ -31,7 +31,7 @@ babylon-mmd is a library focused on loading PMX models and VMD motion files into
 ## Example
 
 ![Playground example of a MMD](/img/extensions/mmdLoader/PGScreenshot.png)
-<Playground id="#028YR6#18" title="Complete MMD Example" description="Example of a MMD model with a VMD motion file and audio." />
+<Playground id="#028YR6#41" title="Complete MMD Example" description="Example of a MMD model with a VMD motion file and audio." />
```
Popov72, line 34: There's a an error in the shader when testing https://playground.babylonjs.com/#028YR6#41, not sure it is expected (the PG seems to work, though)?

image

RaananW: There is also an small issue when running the playground again (the model starts playing while a new one being loaded), and when going to and out of fullscreen (controls are invisible)

noname0310: Shader compilation error has now been resolved. The outline renderer was not working in the Mirror reflection render target texture.

image
@RaananW 
As for the problem with the control, are you referring to the problem caused by the document coming out of the bottom of the iframe? I didn't quite understand that.

RaananW: I am referring to the playgrounditself

noname0310: It seems to be caused by the next scene being loaded before all the async operations in CreateScene are finished.
I was able to reproduce the problem easily by quickly spamming the play button.

Is it possible to make the playground wait for the asynchronized scene creation?

If you think this is a topic worth discussing, I'll post it in the forum

RaananW: This is not entirely related to this, though it is a viable concern.

When I opened the playground yesterday, waiting for it to load, entered fullscreen and quit it, and ran the playground again, it failed to load correctly and failed to show the control area. This seems to be fixed right now.

RaananW (approved):

PR #1144, opened 2024-06-04
Update theFirstStep.md

MelisaIm:
While following the steps, removing the diffuse color line results in a different outcome than the red/black checkerboard expected. Changing instruction to add the new lines instead of replace.

What I expected as a new learner: 
Screenshot 2024-06-03 at 8 50 44 PM

What I saw after following the instructions as presented: 
Screenshot 2024-06-03 at 8 51 51 PM


RaananW (approved):
Fits to the playgrounds as well. thanks a lot for this!

PR #1149, opened 2024-06-12
Fix pg for filtering in collisionEvents.md

BarthPaleologue:
- The mask should use a binary "or" instead of "and"
- Apply the filter to sphere2 and sphere1 instead of sphere1 every time


CedricGuillemet (approved):
Good catch!

PR #1152, opened 2024-06-20
Preparation to move the docs to the main repository

RaananW:
This removes a dependency that prevents us from using npm on this repository.

It is still using yarn, but npm is fully supported.

tpedoc was also updated to support a newer version of typescript.


RaananW:
Part of https://github.com/BabylonJS/Babylon.js/issues/15094

sebavan (approved):

PR #1153, opened 2024-06-23
Update es6Support.md - Fix npm package link mistake

leohumnew:
Fix the Viewer npm package link mistakenly pointing to the Inspector package, and add the Inspector package on its own line.


RaananW (approved):
thanks! man, how did we miss that...

PR #1158, opened 2024-06-29
Html parsing in <Alert> and GUI "introduction"/"how to use it" change

lepetitglacon:


RaananW (approved):
I don't like the unsafe innerHTML set, however - i like the change which makes perfect sense.

PR #1182, opened 2024-08-28
Update mesh picking doc

PatrickRyanMS:
Add information about picking directly with Ray class and how the Ray needs to be transformed to local space of the picked mesh to determine the correct picked point on the mesh.

@CedricGuillemet, this is an update in response to the [forum post](https://forum.babylonjs.com/t/why-scene-pickwithray-mesh-intersects/52941/3) about the Ray class and local vs. world coordinates when picking a mesh. Let me know if this makes sense and covers the question accurately. Thanks!


sebavan (approved):

CedricGuillemet (approved):
Perfect!

PR #1256, opened 2025-01-07
Replace raw code with built-in function

e4779:
Hi! While studying the documentation and source code, I found that the code was taken from a function that is already [written](https://github.com/BabylonJS/Babylon.js/blob/e2b436066bcc2698ba2c72d5f3d19346223522ed/packages/dev/core/src/Misc/tools.ts#L953) and used, for example, [in Devtools](https://github.com/BabylonJS/Babylon.js/blob/e2b436066bcc2698ba2c72d5f3d19346223522ed/packages/dev/inspector/src/components/actionTabs/tabs/toolsTabComponent.tsx#L305).

Maybe it would be easier and more straightforward to use it?

### DIFF
--- a/content/features/featuresDeepDive/Exporters/Save_Babylon.md
+++ b/content/features/featuresDeepDive/Exporters/Save_Babylon.md
@@ -31,15 +31,7 @@ function doDownload(filename, scene) {
 
   const blob = new Blob([strScene], { type: "octet/stream" });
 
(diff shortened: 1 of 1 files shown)


bjsplat:

https://snapshots-cvgtc2eugrd3cgfd.z01.azurefd.net/DOCS/refs/pull/1256/merge

Popov72 (approved):

sebavan (approved):

PR #1267, opened 2025-01-30
Added initial documentation to Area Lights

SergioRZMasson:
Added initial documentation for Area Lights.

### DIFF
--- a/content/features/featuresDeepDive/lights/lights_introduction.md
+++ b/content/features/featuresDeepDive/lights/lights_introduction.md
@@ -90,7 +90,7 @@ In these playgrounds see how the specular color (green) is combined with the dif
 
 <Playground id="#20OAV9#5" title="Hemispheric Light Example" description="Simple Example of adding a Hemispheric Light to your scene." image="/img/playgroundsAndNMEs/divingDeeperLightsIntro4.jpg" isMain={true} category="Lights"/>
 
-<Playground id="#T7FXR8#11" title="Rectangular Area Light Example" description="Simple Example of adding a Rectangular Area Light to your scene." image="/img/playgroundsAndNMEs/areaLightStandardExample.jpg" isMain={true} category="Lights"/>
+<Playground id="#T7FXR8#20" title="Rectangular Area Light Example" description="Simple Example of adding a Rectangular Area Light to your scene." image="/img/playgroundsAndNMEs/areaLightStandardExample.jpg" isMain={true} category="Lights"/>
 
 For a hemispheric light the _groundColor_ is the light in the opposite direction to the one specified during creation.
 You can think of the _diffuse_ and _specular_ light as coming from the centre of the object in the given direction and the _groundColor_ light in the opposite direction.
--- a/content/features/featuresDeepDive/lights/lights_introduction.md
+++ b/content/features/featuresDeepDive/lights/lights_introduction.md
@@ -63,8 +63,7 @@ const light = new BABYLON.HemisphericLight("HemiLight", new BABYLON.Vector3(0, 1
 
 ### The Rectangular Area Light
 
-A rectangular area light is defined by a width and a height and emits light from the resulting surface towards the -Z direction. Even though the RectAreaLight class itself does not have a direction component, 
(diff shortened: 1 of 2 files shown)


bjsplat:

https://snapshots-cvgtc2eugrd3cgfd.z01.azurefd.net/DOCS/refs/pull/1267/merge

bjsplat:

https://snapshots-cvgtc2eugrd3cgfd.z01.azurefd.net/DOCS/refs/pull/1267/merge

content/features/featuresDeepDive/lights/lights_introduction.md
```diff
@@ -61,6 +61,20 @@ that the full effect is achieved.
 const light = new BABYLON.HemisphericLight("HemiLight", new BABYLON.Vector3(0, 1, 0), scene);
 ```
 
+### The Rectangular Area Light
+
+A rectangular area light is defined by its position, width, and height. It emits light from the resulting surface towards the -Z direction. Even though the RectAreaLight class itself does not have a direction component, it can be attached to a transform node to be rotated and moved around.
+
+_A simple use of a rectangular area light_
```
PatrickRyanMS: Are we missing an image here? The "simple use of a rectangular area light" seems to imply there should be an image.

content/features/featuresDeepDive/lights/lights_introduction.md
```diff
@@ -61,6 +61,20 @@ that the full effect is achieved.
...
+
+A rectangular area light is defined by its position, width, and height. It emits light from the resulting surface towards the -Z direction. Even though the RectAreaLight class itself does not have a direction component, it can be attached to a transform node to be rotated and moved around.
+
+_A simple use of a rectangular area light_
+
+```javascript
+const light = new BABYLON.RectAreaLight("areaLight", new BABYLON.Vector3(0, 1, 0), 2, 2, scene);
+```
+
+One important thing to note is that, due to differences in implementation, StandardMaterial will reflect light based on its "roughness" value (instead of specular power, as with other lights). This difference in behavior is significant, and the scene should set the correct roughness value to achieve the desired look.
```
PatrickRyanMS, line 72: I think this section could use a little more context. You mentioned the difference in the StandardMaterial roughness parameter, and mentioned that the scene should set this correctly. It seems like this could be expanded to describe if this process is completely automatic or if the user has any control over the roughness parameter. If the area light does not look right for specular reflections, is there anything to do to fix this value?

content/features/featuresDeepDive/lights/lights_introduction.md
```diff
@@ -78,6 +92,8 @@ In these playgrounds see how the specular color (green) is combined with the dif
 
 <Playground id="#20OAV9#5" title="Hemispheric Light Example" description="Simple Example of adding a Hemispheric Light to your scene." image="/img/playgroundsAndNMEs/divingDeeperLightsIntro4.jpg" isMain={true} category="Lights"/>
 
+<Playground id="#T7FXR8#11" title="Rectangular Area Light Example" description="Simple Example of adding a Rectangular Area Light to your scene." image="/img/playgroundsAndNMEs/areaLightStandardExample.jpg" isMain={true} category="Lights"/>
```
PatrickRyanMS: The example has a fly camera in it, which is a little hard to use to move the scene around to see how the area lights are illuminating and reflecting in the floor. I think an ArcRotate would be easier to use.

content/features/featuresDeepDive/lights/lights_introduction.md
SergioRZMasson: Users will need to set roughness manually. However, we do have code in the exporters that converts specularPower to roughness and exporting StandardMaterial to GLTF. I have a PR in Babylon.js that tries to export that so users have access to it. When that PR gets merged I will update this documentation showing how to use that feature.

bjsplat:

https://snapshots-cvgtc2eugrd3cgfd.z01.azurefd.net/DOCS/refs/pull/1267/merge

sebavan (approved):

PR #1330, opened 2025-04-07
Fixed ES6 imports and improved FAQ

mlyon-df:
Some of the code didn't work as written due to incorrect import statements. I also took the opportunity to make a couple other improvements.

Changes:
- Add `.js` extension to legacy `BABYLON` import
- Add `.js` extension to imports in 'Creating our first js APP' section
- Change 6 uses of `var` to `const` in 'Creating our first js APP' section
- Add a note to 'Creating our first js APP' explaining the use of files instead of base modules
- Add `.js` extension to imports in 'Typescript' section
- Change 6 uses of `var` to `const` in 'Typescript' section
- Add a note to 'Typescript' section explaining the use of files instead of base modules
- Add `.js` to several imports in the FAQ section
- Fix the import statement in 'How do I efficiently use the `Mesh.Create...` methods?' FAQ to properly import CreateBox
- Fix the import statement in 'How does deserialization work?' FAQ to properly import PBRMaterial
- Update 'The intellisense does not propose...' FAQ to call out that the problem is IDE related
- Add 'Why do I have a TypeError...' FAQ to help when intellisense offers methods that aren't yet available

### DIFF
--- a/content/setup/frameworkPackages/es6Support.md
+++ b/content/setup/frameworkPackages/es6Support.md
@@ -353,14 +353,14 @@ The treatment even if a bit annoying is simple: you need to import manually only
 
 _How do I efficiently use the `Mesh.Create...` methods?_
 
-The simplest is to load only the builder corresponding to your construction method. If you wish to use the `CreateBox` method, you can simply `import { CreateBox } from "@babylonjs/core/Meshes/Builders/boxBuilder.js";` to ensure that the dependant modules have been loaded.**Except if you are relying on all the MeshBuilder methods, we would recommend to not use it directly but favor the smaller builders**.
+The simplest is to load only the builder corresponding to your construction method. If you wish to use the `CreateBox` method, you can simply `import "@babylonjs/core/Meshes/Builders/boxBuilder.js";` to ensure that the dependant modules have been loaded.**Except if you are relying on all the MeshBuilder methods, we would recommend to not use it directly but favor the smaller builders**.
 
 _Why using the default material is not working?_
 By default, any mesh in a scene are using the scene defaultMaterial. With tree shaking you might not need this material so we do not force it as a dependency in the code. That said, would you need to use it, you can simply `` `import "@babylonjs/core/Materials/standardMaterial.js";` `` to ensure that the default material would be operational.
 
 _How does deserialization work?_
 
-When you deserialize a Babylon.js object like a Material or Light, it is impossible for the framework to know beforehand what kind of entity is enclosed in your file. For instance, are you relying on Standard vs PBRMaterial. We again rely on side effect here and the deserialization will only be able to load the kind of entity you have imported in your app. This means if you know you will need to deserialize a PBRMaterial, you can `import { PBRMaterial } from "@babylonjs/core/Materials/PBR/pbrMaterial.js";` beforehand.
+When you deserialize a Babylon.js object like a Material or Light, it is impossible for the framework to know beforehand what kind of entity is enclosed in your file. For instance, are you relying on Standard vs PBRMaterial. We again rely on side effect here and the deserialization will only be able to load the kind of entity you have imported in your app. This means if you know you will need to deserialize a PBRMaterial, you can `import "@babylonjs/core/Materials/PBR/pbrMaterial.js";` beforehand.
 
 _How do I know if I am importing a folder or a file?_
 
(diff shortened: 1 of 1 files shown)


bjsplat:

https://snapshots-cvgtc2eugrd3cgfd.z01.azurefd.net/DOCS/refs/pull/1330/merge

content/setup/frameworkPackages/es6Support.md
```diff
@@ -349,14 +353,14 @@ The treatment even if a bit annoying is simple: you need to import manually only
 
 _How do I efficiently use the `Mesh.Create...` methods?_
 
-The simplest is to load only the builder corresponding to your construction method. If you wish to use the `CreateBox` method, you can simply `import "@babylonjs/core/Meshes/Builders/boxBuilder";` to ensure that the dependant modules have been loaded.**Except if you are relying on all the MeshBuilder methods, we would recommend to not use it directly but favor the smaller builders**.
+The simplest is to load only the builder corresponding to your construction method. If you wish to use the `CreateBox` method, you can simply `import { CreateBox } from "@babylonjs/core/Meshes/Builders/boxBuilder.js";` to ensure that the dependant modules have been loaded.**Except if you are relying on all the MeshBuilder methods, we would recommend to not use it directly but favor the smaller builders**.
```
RaananW: This line refers to the Mesh.Create??? methods. The `import @.....` is loading the side effects needed to use the Mesh.CreateBox function. So the change is incorrect in this specific context.

content/setup/frameworkPackages/es6Support.md
```diff
@@ -349,14 +353,14 @@ The treatment even if a bit annoying is simple: you need to import manually only
...
 
 _How does deserialization work?_
 
-When you deserialize a Babylon.js object like a Material or Light, it is impossible for the framework to know beforehand what kind of entity is enclosed in your file. For instance, are you relying on Standard vs PBRMaterial. We again rely on side effect here and the deserialization will only be able to load the kind of entity you have imported in your app. This means if you know you will need to deserialize a PBRMaterial, you can `import "@babylonjs/core/Materials/PBR/pbrMaterial";` beforehand.
+When you deserialize a Babylon.js object like a Material or Light, it is impossible for the framework to know beforehand what kind of entity is enclosed in your file. For instance, are you relying on Standard vs PBRMaterial. We again rely on side effect here and the deserialization will only be able to load the kind of entity you have imported in your app. This means if you know you will need to deserialize a PBRMaterial, you can `import { PBRMaterial } from "@babylonjs/core/Materials/PBR/pbrMaterial.js";` beforehand.
```
RaananW: Same as the last comment. this is more related to the side effects and not the import of PBRMaterial.

RaananW (commented):
I have no issue with the changes in the PR, in general. Two comments I added are regarding the side effects vs. importing a specific part of a module.

content/setup/frameworkPackages/es6Support.md
mlyon-df: I made a best guess on the true intent of this entry and it sounds like I missed the mark. The `Mesh.Create...` methods are marked as deprecated in favor of `MeshBuilder` and I interpreted the last line to be encouraging people to go directly to the function (i.e. `CreateBox`).

Given your comment and the deprecation I just mentioned, I think this answer needs a bit of a different rewrite. I will amend this commit to just be a corrected side effect import and wait until this PR is sorted to suggest the new wording.

RaananW: The question is "_How do I efficiently use the `Mesh.Create...` methods?_", and the answer is - with side effects. it is true that they are deprecated and maybe we can consider removing or making the changes, but this is something we need to discuss in the team. Deprecated in Babylon does not mean that it will ever be removed - we are always back-compat.

content/setup/frameworkPackages/es6Support.md
mlyon-df: I'm not surprised by this. I attempted to read over the code and test the side effects import, but I didn't have a serialized PBRMaterial to do a proper test with. I will adjust.

content/setup/frameworkPackages/es6Support.md
mlyon-df: Fair enough. My thought was to make the answer include information about the deprecated status, give people direction to move to MeshBuilder, and provide information for both approaches. Another reason for this to be handled as a separate commit.

mlyon-df:
@RaananW I have issued commit 8cff814 to cover the discussed changes. I will provide my proposed rewording to the `Mesh.Create...` topic at a future date.

bjsplat:

https://snapshots-cvgtc2eugrd3cgfd.z01.azurefd.net/DOCS/refs/pull/1330/merge

RaananW (approved):

PR #1335, opened 2025-04-11
Update node geometry doc with patterns to wait for build

alexchuber:
The geometry may not be immediately available after `build()`, so we should mention it in the docs + guard against that in our examples.

### DIFF
--- a/content/features/featuresDeepDive/mesh/nodeGeometry.md
+++ b/content/features/featuresDeepDive/mesh/nodeGeometry.md
@@ -68,9 +68,13 @@ The `GeometryOutputBlock` will collect the final VertexData that can then be use
 
 ```
 // Build and instantiate mesh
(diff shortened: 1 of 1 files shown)


bjsplat:

https://snapshots-cvgtc2eugrd3cgfd.z01.azurefd.net/DOCS/refs/pull/1335/merge

RaananW (commented):
If that's the case, is build not async?

RaananW:
Ok, looked at the code. It should be async IMO, but isn't. if there are promises involved, it waits for these promises (while already returning void) and the runs build again. This is why it is working.

RaananW (approved):
API should be promise-based, but is void. this solution works correctly.

alexchuber:
Yeah, I personally agree that it should be promise-based (even just via a `buildAsync()` method), but I'll leave that decision up to @deltakosh when he's back.

PR #1345, opened 2025-05-06
WebGPU NDC is different from WebGL's

lockphase:
Inform user about NDC difference in glsl vs wgsl, especially for the z-axis as it can foul up shader results.


bjsplat:

https://snapshots-cvgtc2eugrd3cgfd.z01.azurefd.net/DOCS/refs/pull/1345/merge

Popov72 (approved):

PR #1402, opened 2025-09-24
Updating PG search to handle PG 2 manifest

VicenteCartas:
This PR updates the Playground search page to handle the new manifest that the Playground V2 uses (from this PR: https://github.com/BabylonJS/Babylon.js/pull/17175).

The schema of the manifest of the PG V2 is defined as:
```
export type V2Manifest = {
    v: 2;
    language: "JS" | "TS";
    entry: string;
    imports: Record<string, string>;
    files: Record<string, string>;
    cdnBase?: string;
};
```

For search, we care about the _files_ property, which contains pairs with the name of the file and the code of the file. Most of the parsing changes are related to dealing with the fact that now there are multiple files to search in a PG and to surface the name of the file where the code is found (if we are searching by Code).

All search behavior is identical for V1 and V2, except for the following changes: 
- If the playground is a V2 playground, when searching for code with an exact match the name of the file will be printed with the results:
image
- Ellipsis in the code are removed for long lines, this is done to simplify the code that handles finding the correct line and the current page handles long lines correctly by wrapping the lines.

### DIFF
--- a/components/contentComponents/playgroundSearchResult.tsx
+++ b/components/contentComponents/playgroundSearchResult.tsx
@@ -28,6 +28,7 @@ export interface IPlaygroundSearchResult {
 
 const MaxLineLength = 200;
 const NumberOfLinesToShow = 10;
+const CodeNotFound = -1;
 
 export const PlaygroundSearchResult: FunctionComponent<{ searchResult: IPlaygroundSearchResult; type: SearchType; term?: string; setActiveExample: (example: IExampleLink) => void }> = ({ searchResult, type, term, setActiveExample }) => {
     const theme = useTheme();
@@ -102,7 +103,7 @@ export const PlaygroundSearchResult: FunctionComponent<{ searchResult: IPlaygrou
 
                 // When not searching for code, just show the beginning of the first file
                 if (type !== "code") {
-                    TrimLinesAndShow(filesCode.length ? filesCode[0][1].split("\n") : [], 0, -1);
+                    TrimLinesAndShow(filesCode.length ? filesCode[0][1].split("\n") : [], 0, CodeNotFound);
                     return;
                 }
 
@@ -123,7 +124,7 @@ export const PlaygroundSearchResult: FunctionComponent<{ searchResult: IPlaygrou
                     for (foundLine = 0; foundLine < codeLines.length; ++foundLine) {
                         if (codeLines[foundLine].toLowerCase().indexOf(lowerTerm) !== -1) {
                             codeFound = true;
-                            startingLine = Math.max(foundLine - 5, 0);
+                            startingLine = Math.max(foundLine - Math.round(NumberOfLinesToShow / 2), 0);
                             foundFile = filename;
                             break;
                         }
@@ -143,7 +144,7 @@ export const PlaygroundSearchResult: FunctionComponent<{ searchResult: IPlaygrou
                 } else {
                     // Couldn't find the exact code in the snippet, but search thought it was relevant,
                     // so we still show the beginning of the first file
-                    TrimLinesAndShow(filesCode.length ? filesCode[0][1].split("\n") : [], 0, -1);
+                    TrimLinesAndShow(filesCode.length ? filesCode[0][1].split("\n") : [], 0, CodeNotFound);
                 }
             });
         });
--- a/components/contentComponents/playgroundSearchResult.tsx
+++ b/components/contentComponents/playgroundSearchResult.tsx
@@ -26,6 +26,9 @@ export interface IPlaygroundSearchResult {
     version: number;
 }
 
+const MaxLineLength = 200;
+const NumberOfLinesToShow = 10;
+
 export const PlaygroundSearchResult: FunctionComponent<{ searchResult: IPlaygroundSearchResult; type: SearchType; term?: string; setActiveExample: (example: IExampleLink) => void }> = ({ searchResult, type, term, setActiveExample }) => {
     const theme = useTheme();
     const [expanded, setExpanded] = useState<boolean>(false);
@@ -53,6 +56,28 @@ export const PlaygroundSearchResult: FunctionComponent<{ searchResult: IPlaygrou
         // get the snippet code
         fetch(`https://snippet.babylonjs.com/${searchResult.snippetIdentifier}/${searchResult.version}`).then((response) => {
             response.text().then((text) => {
+                const TrimLinesAndShow = (allLines: string[], start: number, found: number) => {
+                    // Clamp indexes
+                    const clampedStart = Math.max(0, start);
+                    const clampedEnd = Math.min(clampedStart + NumberOfLinesToShow, allLines.length);
+
+                    // Only pass to the highlighter what we need to show
+                    const visible: string[] = [];
+                    for (let i = clampedStart; i < clampedEnd; ++i) {
+                        let line = allLines[i] ?? "";
+                        if (line.length > MaxLineLength) {
+                            line = line.substring(0, MaxLineLength) + " ...";
+                        }
+                        visible.push(line);
+                    }
+
+                    setCodeToShow({
+                        code: visible.join("\n"),
+                        startingLine: clampedStart,
+                        foundLine: found,
+                    });
+                };
+
                 const parsed = JSON.parse(JSON.parse(text).jsonPayload);
                 let filesCode: [string, string][] = [];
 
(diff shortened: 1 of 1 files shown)


bjsplat:

https://snapshots-cvgtc2eugrd3cgfd.z01.azurefd.net/DOCS/refs/pull/1402/merge

bjsplat:

https://snapshots-cvgtc2eugrd3cgfd.z01.azurefd.net/DOCS/refs/pull/1402/merge

Popov72 (approved):

Popov72:
I wonder if we can't just do `(V2Manifest.v ?? 1) >= 2` to check if the snippet is a v2 snippet or not?

VicenteCartas:
> I wonder if we can't just do `(V2Manifest.v ?? 1) >= 2` to check if the snippet is a v2 snippet or not?

This field doesn't exist at all in v1, it's just the code as a string. We will be able to use that version in the future if we have other PG updates.

Popov72:
> > I wonder if we can't just do `(V2Manifest.v ?? 1) >= 2` to check if the snippet is a v2 snippet or not?
> 
> This field doesn't exist at all in v1, it's just the code as a string. We will be able to use that version in the future if we have other PG updates.

If `V2Manifest` is just a string in version 1, then `(V2Manifest.v ?? 1) >= 2` should work (because `V2Manifest.v` will be **undefined**)? I don't know how long it takes to execute `parseJSON(code)` when `code` is not a json-stringified code, so maybe it's not worth changing?

bjsplat:

https://snapshots-cvgtc2eugrd3cgfd.z01.azurefd.net/DOCS/refs/pull/1402/merge

VicenteCartas:
> > > I wonder if we can't just do `(V2Manifest.v ?? 1) >= 2` to check if the snippet is a v2 snippet or not?
> > 
> > 
> > This field doesn't exist at all in v1, it's just the code as a string. We will be able to use that version in the future if we have other PG updates.
> 
> If `V2Manifest` is just a string in version 1, then `(V2Manifest.v ?? 1) >= 2` should work (because `V2Manifest.v` will be **undefined**)? I don't know how long it takes to execute `parseJSON(code)` when `code` is not a json-stringified code, so maybe it's not worth changing?

You are right, and that is what the PG V2 PR is doing for the visual tests. But when testing, I found some PGs that were saved with the V2 format, but without .v, so I didn't know how to handle those without doing the parse. Let me think a little about this, I don't like using an exception for control flow but it seemed the easier way to handle this. I'll give it another thought.

sebavan:
Converting to draft to prevent accidental merge before you finalize it. Ping me once done.

bjsplat:

https://snapshots-cvgtc2eugrd3cgfd.z01.azurefd.net/DOCS/refs/pull/1402/merge

Popov72:
> > > > I wonder if we can't just do `(V2Manifest.v ?? 1) >= 2` to check if the snippet is a v2 snippet or not?
> > > 
> > > 
> > > This field doesn't exist at all in v1, it's just the code as a string. We will be able to use that version in the future if we have other PG updates.
> > 
> > 
> > If `V2Manifest` is just a string in version 1, then `(V2Manifest.v ?? 1) >= 2` should work (because `V2Manifest.v` will be **undefined**)? I don't know how long it takes to execute `parseJSON(code)` when `code` is not a json-stringified code, so maybe it's not worth changing?
> 
> You are right, and that is what the PG V2 PR is doing for the visual tests. But when testing, I found some PGs that were saved with the V2 format, but without .v, so I didn't know how to handle those without doing the parse. Let me think a little about this, I don't like using an exception for control flow but it seemed the easier way to handle this. I'll give it another thought.

Maybe you can check the type: if it is "string", it is v1, else it is v2?

components/contentComponents/playgroundSearchResult.tsx
```diff
@@ -54,27 +54,87 @@ export const PlaygroundSearchResult: FunctionComponent<{ searchResult: IPlaygrou
...
+
+                for (const [filename, code] of filesCode) {
+                    codeFound = false;
+                    codeLines = (code || "").split("\n");
+                    startingLine = 0;
+                    foundLine = -1;
+                    const lowerTerm = term.toLowerCase();
+                    
                     for (foundLine = 0; foundLine < codeLines.length; ++foundLine) {
-                        if (codeLines[foundLine].length > 100) {
```
alexchuber, line 64: I would put this check back; it doesn't hurt. There are a ton of playgrounds that have, for example, entire base64-encoded buffer data in a single line. The chances of someone searching for a term past 100 chars feels slim.

alexchuber: Actually, I guess I'm not sure what takes longer.... making a substring of a giant line, or searching a giant line.

VicenteCartas: The page already wraps if a line is too long the correct way, so I felt we do not need the elipsis anymore. It feels strange on a wide monitor to have the line cut to the ... instead of showing the whole thing. Happy to add it back if we think that is better.

components/contentComponents/playgroundSearchResult.tsx
```diff
@@ -54,27 +54,87 @@ export const PlaygroundSearchResult: FunctionComponent<{ searchResult: IPlaygrou
...
+                        v2Manifest.files.default = parsed.code;
+                    }
+                } catch (e) {
+                    // If parsing the JSON failed, treat it as code just in case
+                    v2Manifest.files.default = parsed.code;
+                }
+
+                // Put the files and their code into a list for easier access
+                for (const [filename, code] of Object.entries(v2Manifest.files)) {
+                    if (code && typeof code === 'string') {
```
bghgary: Why do we need to check `code` but not `filename`? Seems like we should either check both or check none.

components/contentComponents/playgroundSearchResult.tsx
```diff
@@ -54,27 +54,87 @@ export const PlaygroundSearchResult: FunctionComponent<{ searchResult: IPlaygrou
...
                     }
+
+                    if (codeFound) {
+                        break;
+                    }
+                }
+                
+                if (codeFound) {
+                    if (foundFile !== "default") {
+                        setName(name + ` - ${foundFile}`);
```
bghgary: nit:
```suggestion
                        setName(`${name} - ${foundFile}`);
```

components/contentComponents/playgroundSearchResult.tsx
```diff
@@ -54,27 +54,87 @@ export const PlaygroundSearchResult: FunctionComponent<{ searchResult: IPlaygrou
...
+                    setCodeToShow({
+                        code: codeLines.slice(startingLine, startingLine + 10).join("\n"),
+                        startingLine,
+                        foundLine,
+                    });
+                } else {
+                    // Couldn't find the exact code in the snippet, but search thought it was relevant,
+                    // so we still show the first 10 lines of the first file
+                    setCodeToShow({
+                        code: filesCode.length ? filesCode[0][1].split("\n").slice(0, 10).join("\n") : "",
```
bghgary: nit: maybe can factor out to a small function since it's used at least twice

bghgary (approved):
Just some nits mostly

VicenteCartas:
> > > > > I wonder if we can't just do `(V2Manifest.v ?? 1) >= 2` to check if the snippet is a v2 snippet or not?
> > > > 
> > > > 
> > > > This field doesn't exist at all in v1, it's just the code as a string. We will be able to use that version in the future if we have other PG updates.
> > > 
> > > 
> > > If `V2Manifest` is just a string in version 1, then `(V2Manifest.v ?? 1) >= 2` should work (because `V2Manifest.v` will be **undefined**)? I don't know how long it takes to execute `parseJSON(code)` when `code` is not a json-stringified code, so maybe it's not worth changing?
> > 
> > 
> > You are right, and that is what the PG V2 PR is doing for the visual tests. But when testing, I found some PGs that were saved with the V2 format, but without .v, so I didn't know how to handle those without doing the parse. Let me think a little about this, I don't like using an exception for control flow but it seemed the easier way to handle this. I'll give it another thought.
> 
> Maybe you can check the type: if it is "string", it is v1, else it is v2?

They are both strings, I could check if the first character is { and assume that's a json object in string form, but I think this is good with checking for the version.

components/contentComponents/playgroundSearchResult.tsx
VicenteCartas: Re-added at 200 characters max, that will show more code in wide monitors and avoid the slowdown of base64 strings.

bjsplat:

https://snapshots-cvgtc2eugrd3cgfd.z01.azurefd.net/DOCS/refs/pull/1402/merge

bjsplat:

https://snapshots-cvgtc2eugrd3cgfd.z01.azurefd.net/DOCS/refs/pull/1402/merge

alexchuber (approved):

PR #1448, opened 2025-12-12
Fix audio engine live synth playground

docEdub:
Dynamic imports don't work in the new playground, so the existing audio engine live synth playground does not work anymore. This change updates the playground to a version that works in the playground.


Popov72 (approved):

bjsplat:

https://snapshots-cvgtc2eugrd3cgfd.z01.azurefd.net/DOCS/refs/pull/1448/merge

PR #1457, opened 2025-12-22
docs: correct SceneLoader.Append to separate rootUrl and filename par…

EANadim:
…ameters

The example was missing the separate filename parameter, causing the scene object to be misinterpreted as the filename. Split the URL into rootUrl and sceneFilename parameters as required by the API. Also, please correct the playground example so that it works


bjsplat:

https://snapshots-cvgtc2eugrd3cgfd.z01.azurefd.net/DOCS/refs/pull/1457/merge

Popov72 (approved):

PR #1544, opened 2026-03-13
Fix shadow PG

Popov72:
See https://forum.babylonjs.com/t/shadowdepthwrapper-getting-error-with-an-async-adding-with-standalone-true/62769


bjsplat:

https://snapshots-cvgtc2eugrd3cgfd.z01.azurefd.net/DOCS/refs/pull/1544/merge

PR #1559, opened 2026-04-01
security: pin all dependencies to exact versions

RaananW:
## Summary

Pin all external dependencies in `package.json` to their exact resolved versions from `package-lock.json` and add an `.npmrc` with `min-release-age=2d`.

## Changes

- **package.json**: Removed `^` and `~` prefixes from all external dependencies (both `dependencies` and `devDependencies`), pinning them to the exact versions currently in `package-lock.json`
- **package.json**: Updated `resolutions` field to match the pinned `@types/react` version
- **.npmrc**: Added `min-release-age=2d` to reject any package version published less than 2 days ago

## Motivation

Prevent unexpected version drift and reduce supply-chain attack surface by:
1. Ensuring `npm install` always installs the exact audited versions
2. Rejecting very recently published packages that haven't had time for community review


bjsplat:

https://snapshots-cvgtc2eugrd3cgfd.z01.azurefd.net/DOCS/refs/pull/1559/merge

.npmrc
```diff
@@ -0,0 +1 @@
+min-release-age=2d
```
sebavan, line 1: could it not prevent security fixes ?

PR #1580, opened 2026-05-20
Modernize documentation build architecture

RaananW:
## Summary

This draft PR modernizes the documentation site architecture while preserving the existing static export deployment model. It introduces a clearer separation between markdown compilation, typed content discovery, validation, static artifact generation, TypeDoc generation, docs-page UI state, and future App Router migration planning.

The goal is to make the documentation build easier to reason about, easier to validate, and less dependent on route-level side effects, without changing the server architecture or requiring SSR/ISR.

## Why this is better and more stable

- Keeps `output: "export"` as the deployment contract and validates that the site still builds as static HTML.
- Centralizes markdown compilation so home/docs/tests use the same pipeline instead of subtly different parsing paths.
- Adds a typed content graph as the source of truth for route manifests, breadcrumbs, examples, and navigation data.
- Adds a content validation gate before build output is generated, catching broken links, invalid structure entries, frontmatter issues, and incomplete example metadata earlier.
- Makes generated artifacts explicit through dedicated scripts for content, sitemap, search data, example preview images, TypeDoc, and App Router evaluation.
- Moves TypeDoc generation out of route utilities, so pages read prebuilt artifacts instead of mixing generation logic with static path discovery.
- Isolates docs-page UI concerns under `features/docs`, reducing the size and responsibility of the catch-all docs route.
- Evaluates App Router migration readiness without adding conflicting `app/` routes in this branch.

## Main changes

- Added shared markdown helpers under `lib/markdown`.
- Added typed content graph, static artifact, example image, and validation utilities under `lib/contentGraph`.
- Added explicit build scripts:
  - `build:content`
  - `validate:content`
  - `build:sitemap`
  - `build:search`
  - `build:typedoc`
  - `build:example-images`
  - `check:example-images`
  - `evaluate:app-router`
- Refactored docs rendering into `features/docs` providers/components/hooks.
- Split TypeDoc route-safe artifact readers from build-only TypeDoc generation utilities.
- Added regression tests for the content graph, artifact generation, validation, docs UI feature helpers, TypeDoc artifacts, and App Router evaluation.
- Added `documentation-modernization-plan.md` as the shared implementation and verification record.

## App Router evaluation result

The static data pipeline is now ready for a dedicated App Router proof branch, but this PR intentionally does not migrate production routes to App Router.

[...]

PR #1606, opened 2026-08-07
Restore API pages in the search index

RaananW:
## Problem

API (TypeDoc) pages stopped appearing in the site search. The "API Only" filter on `/search` returns nothing, and API results never show up in general searches.

The regression came from bf967a6d ("Modernize documentation build architecture", #1580). Before that refactor, `getAPIPageData()` indexed each API page as a side effect while Next.js statically rendered it:

```ts
addSearchItem({ isApi: true, content: htmlToText(html), ... });
```

The refactor moved search index generation into `scripts/build-content.ts` / `scripts/build-search.ts`, which derive records purely from the markdown content graph — and `createDocumentationSearchIndex` hardcodes `isApi: false`. After the refactor, nothing in the codebase produced an `isApi: true` record.

## Fix

Adds a dedicated build step that generates API search records from the already-generated TypeDoc HTML, instead of relying on page-render side effects.

- **`lib/buildUtils/typedocSearchIndex.utils.ts`** (new) — walks the generated TypeDoc output, extracts a title/description/content per page with `html-to-text`, and emits `ISearchIndexItem[]` with `isApi: true`. Covers all three targets: `typedoc` and `packages/viewer` (babylon flavor) and `lite/typedoc` (lite flavor).
- **`scripts/build-typedoc.ts` / `scripts/build-search.ts`** — call `buildApiSearchIndex()` so the artifact is written alongside the documentation and playground indexes.
- **`scripts/upload-search-index.ts`** — reads the artifact and uploads API items with the rest.

Document ids keep the existing `base64(url)` scheme, so this updates the pre-existing API documents in place via `mergeOrUpload` rather than duplicating them.

### Build-only, not bundled

`html-to-text` is a devDependency and `lib/buildUtils/typedoc.utils.ts` *is* reachable from `pages/typedoc/[[...id]].tsx`. The new code therefore lives in a separate build-only module (mirroring `typedocBuild.utils.ts`) and is imported only from `scripts/` and `__tests__/`.

## Clearing behaviour on master

`upload:search -- --clear` runs only on master. `clearIndex(true, apiPaths, flavor)` prunes API documents whose path is no longer in the freshly generated set, then survivors are updated in place. Clears complete before any upload begins.

Two safety issues surfaced during review and are addressed here:

1. **Partial-index wipe.** `clearIndex` with an empty `doNotDelete` deletes everything matching the filter. A flavor that produced no fresh API records would have had its entire live API index wiped. The artifact now records `indexedTargets`, and the API clear is gated on `isFlavorFullyIndexed()` — every declared target for that flavor must have been indexed.

   This is not theoretical: `isBabylonLiteRepositoryPublic()` swallows all failures (including GitHub's unauthenticated 60/hr rate limit) and returns `false`. `build:content` and `build:typedoc` are separate processes that call it independently, so they can disagree, and the Lite early-return exits 0 — CI would proceed to `upload:search --clear`.

2. **Cross-contamination between clears.** Adding API items to `flavorsToUpload` let a flavor enter the clear loop on API items alone, where the unguarded documentation and playground clears would then wipe live docs/playgrounds. All three clear operations are now independently guarded, and all guards fail closed — worst case is stale documents surviving, never live documents being lost.

Also fixed: the babylon clear filter was `flavor eq 'babylon'`, which never matched documents indexed before the `flavor` field existed, so legacy entries could never be pruned. It is now `flavor eq 'babylon' or flavor eq null`.

## Other changes

- `search.utils.ts` — `addSearchItem`/`addPlaygroundItem` now batch at 100 docs/request. Content is capped at 15,000 chars, keeping batches around 1.5 MB against Azure Search's 16 MB / 1000-doc limits.
- `typedocBuild.utils.ts` — TypeDoc generation failures previously logged and continued, which could mark a target as indexed with partial output. They now remove the partial output directory and rethrow.

Categories are deliberately omitted for API items: `searchResult.component.tsx` builds breadcrumb URLs from `path.split("/")` slices, which would produce broken links like `/typedoc/classes`. Omitting them keeps the breadcrumb row hidden, matching pre-refactor behaviour.

## Validation

- `npx tsc --noEmit` clean; `npm test` passes (80 tests, 8 new).
- Extraction verified against genuine TypeDoc 0.28.20 output rather than hand-written fixtures — the real markup nests `.tsd-comment` inside `.tsd-panel`, which an initial hand-written fixture got wrong.
- End-to-end run of `build:search` against staged TypeDoc output produced correct paths, flavors and `isApi` values across all three targets.
- Dry-ran `upload:search --clear` with a stubbed `fetch` across `complete`, `lite-docs-missing`, `viewer-missing` and `no-api` scenarios, plus a prune scenario mixing current, removed and legacy un-flavored documents (paginated). Stale and legacy entries were deleted, current ones preserved and re-uploaded.

Not run: a full `npm run build:typedoc`, which clones Babylon.js and takes 20-40 minutes.

Co-authored-by: Copilot App <223556219+Copilot@users.noreply.github.com>


bjsplat:

Snapshot: [https://snapshots-cvgtc2eugrd3cgfd.z01.azurefd.net/DOCS/refs/pull/1606/merge](https://snapshots-cvgtc2eugrd3cgfd.z01.azurefd.net/DOCS/refs/pull/1606/merge)
Path: `DOCS/refs/pull/1606/merge`
Build: https://dev.azure.com/babylonjs/ContinousIntegration/_build/results?buildId=57521

sebavan (approved):

PR #1624, opened 2026-09-14
Root motion for animation groups

Pryme8:
Documents the root motion API added in BabylonJS/Babylon.js#18908: `RootMotionClip` and `RootMotionController`.

- New page `content/features/featuresDeepDive/animation/rootMotion.md` in the Animation deep dive: what root motion is, a quick start, how a clip finds the travel, turning, the controller and its blending, driving a physics controller from the deltas, playback behaviors, requirements and limits, and the options table.
- Navigation entry in `configuration/structure.json`, after Animation Retargeting.

Draft until the feature PR is merged. The playground linked from the page (#ZE8Y5A#0) needs that build.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


Pryme8:
I will get this updated and out of draft this evening.

content/features/featuresDeepDive/animation/rootMotion.md
```diff
@@ -0,0 +1,152 @@
...
+controller.applyToCharacter = false;
+controller.onRootMotionObservable.add(() => {
+    hero.moveWithCollisions(controller.deltaPosition);
+});
+```
+
+### Jumps and resets
+
+The motion follows the playback, so starting, stopping, pausing and changing the speed ratio need nothing from you. A jump is different: after `goToFrame`, or after teleporting the character back to the origin, call `controller.reset()` so the jump is not read as motion.
+
```
Popov72: `controller.reset()` is needed when the **animation playback** jumps (for example after `goToFrame` or resetting the group), because the controller tracks runtime-animation progress. Teleporting `characterNode` itself does not change that progress and cannot be read as root motion. Calling `reset()` after an ordinary character teleport instead clears the baseline, so the next evaluated animation step contributes no delta and can introduce a one-step foot slip. Please remove the teleport advice, or clarify that the playback/root animation was jumped too.

Popov72 (changes requested):
The page is thorough and matches the merged implementation closely, but the reset guidance currently conflates an animation-playback jump with moving the character node. That advice would make a normal teleport discard the next root-motion step. Please correct the inline issue below.

Non-blocking: the linked Playground still contains the pre-merge snapshot instructions even though the preview bundle now exposes the root-motion API.

content/features/featuresDeepDive/animation/rootMotion.md
Pryme8: You are right on both counts, and I had the cause wrong in the page: `reset()` forgets where the playbacks were last measured, so a character teleport is none of its business, and calling it anyway drops the next step's travel because the first evaluation after a reset only re-establishes the baseline.

Fixed in 5cc934d. The teleport advice is gone, and the section now says outright that moving the character node yourself needs no reset, with the reason: the controller tracks how far the playback advanced, not where the character is, and resetting without a playback jump costs a step of travel, which is the slip the page exists to avoid.

Pryme8:
Both points are addressed.

**The reset guidance** (5cc934d): you were right, and the page had the cause wrong. `reset()` forgets where the playbacks were last measured, so a character teleport is none of its business, and calling it anyway drops the next step's travel because the first evaluation after a reset only re-establishes the baseline. The teleport advice is gone and the section now says plainly that moving the character node yourself needs no reset, with that reasoning.

**The playground** (1379c4a): the page now links [#ZE8Y5A#1](https://playground.babylonjs.com/#ZE8Y5A#1). It drops the `?snapshot=refs/pull/18908/merge` instruction, the pull request reference in the header comment and the fallback text that pointed at the snapshot build, since the default Playground build carries the API. Version #0 is untouched. Its "Reset position" button no longer resets the controller either, so the demo matches what the page now says.

I ran #ZE8Y5A#1 against the default build before linking it: the character walks at 1.66 m/s, which is the `averageSpeed` the clip reports for that walk, and pressing "Reset position" puts it back at the origin and carries straight on without losing a step.

Popov72 (approved):

bjsplat:

Snapshot: [https://snapshots-cvgtc2eugrd3cgfd.z01.azurefd.net/DOCS/refs/pull/1624/merge](https://snapshots-cvgtc2eugrd3cgfd.z01.azurefd.net/DOCS/refs/pull/1624/merge)
Path: `DOCS/refs/pull/1624/merge`
Build: https://dev.azure.com/babylonjs/ContinousIntegration/_build/results?buildId=59988

PR #1629, opened 2026-09-17
Add a Box3D Physics page to the community extensions

Pryme8:
Adds a community extension page for [babylon-box3d](https://github.com/Pryme8/babylon-box3d), which brings [Box3D](https://github.com/erincatto/box3d) (Erin Catto's 3D successor to Box2D v3, MIT) to Babylon.js as an `IPhysicsEnginePluginV2` implementation. Scenes written for the Havok plugin run on it unchanged: `PhysicsAggregate`, `PhysicsBody`, shapes, constraints, events and ray casts.

The extension is published on npm as `babylon-box3d` and the repository carries the tests and the benchmark harness the page's numbers come from.

The page covers:

- installation, the bundler `locateFile` case, script tags and a complete Playground scene
- an optional threaded build that runs a world step across worker threads on a cross origin isolated page, with the same results at any worker count
- the Box3D specific extras: explosions, wheel joints, parallel joints, collision groups, rolling resistance
- measured step times against Havok and Oimo, with the caveats stated next to them (Havok's own world step is faster in the piles, threads cost more than they save on small scenes, and the pyramid rows are about stability rather than the clock)

`configuration/structure.json` gets the matching entry, alphabetically inside `communityExtensions`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


content/communityExtensions/box3dPhysics.md
```diff
@@ -0,0 +1,219 @@
...
+
+```
+npm add @babylonjs/core babylon-box3d
+```
+
+```typescript
+import { Vector3 } from "@babylonjs/core/Maths/math.vector";
+import { Box3D, Box3DPlugin } from "babylon-box3d";
+
+const box3d = await Box3D();
```
Popov72, line 55: This npm example imports Babylon modules individually but never imports `@babylonjs/core/Physics/joinedPhysicsEngineComponent`. In Babylon.js 9 that side-effect module registers `Scene.enablePhysics`, and `babylon-box3d` does not import it, so this can fail with `scene.enablePhysics is not a function`. Please add `import "@babylonjs/core/Physics/joinedPhysicsEngineComponent";` (or use the `@babylonjs/core` index), as documented in the package README.

content/communityExtensions/box3dPhysics.md
```diff
@@ -0,0 +1,219 @@
...
+The threads are worth asking for on scenes with thousands of awake bodies, where they take a step several times faster than one thread can. Under a few hundred they cost more than they save.
+
+### Script tags
+
+```html
+<script src="https://cdn.babylonjs.com/babylon.js"></script>
+<script src="https://unpkg.com/babylon-box3d/lib/umd/box3d.umd.js"></script>
+<script src="https://unpkg.com/babylon-box3d/umd/babylon.box3d.min.js"></script>
+<script>
+  const box3d = await Box3D();
```
Popov72: This is a classic `<script>`, where top-level `await` is a syntax error. Please wrap initialization in an async function/IIFE, use `Box3D().then(...)`, or make this a valid module-script example.

content/communityExtensions/box3dPhysics.md
```diff
@@ -0,0 +1,219 @@
...
+    url: /features/featuresDeepDive/physics
+  - title: Box3D repository
+    url: https://github.com/erincatto/box3d
+video-overview:
+video-content:
+---
+
+# Box3D Physics
+
+[babylon-box3d](https://github.com/Pryme8/babylon-box3d) brings [Box3D](https://github.com/erincatto/box3d) to Babylon.js. Box3D is Erin Catto's 3D successor to Box2D v3: an MIT licensed rigid body engine with a soft step solver, continuous collision, cross platform determinism and SIMD. The package ships the engine compiled to WebAssembly together with `Box3DPlugin`, an implementation of the physics v2 plugin interface, so everything written for the Havok plugin (`PhysicsAggregate`, `PhysicsBody`, `PhysicsShape`, constraints, events, ray casts) works unchanged.
```
Popov72: “Everything written for the Havok plugin works unchanged” overstates compatibility. This page itself notes that `PhysicsCharacterController` is unsupported and that SixDoF constraints are approximated using the closest Box3D joint; the package documentation also limits mesh shapes to static/animated bodies. Please describe this as support for most Physics V2 APIs and call out the behavioral limitations.

content/communityExtensions/box3dPhysics.md
```diff
@@ -0,0 +1,219 @@
...
+
+  const box3d = await Box3D();
+  const plugin = new BABYLONBOX3D.Box3DPlugin(true, box3d);
+  scene.enablePhysics(new BABYLON.Vector3(0, -9.81, 0), plugin);
+
+  const ground = BABYLON.MeshBuilder.CreateBox("ground", { width: 40, height: 1, depth: 40 }, scene);
+  ground.position.y = -0.5;
+  new BABYLON.PhysicsAggregate(ground, BABYLON.PhysicsShapeType.BOX, { mass: 0, friction: 0.7 }, scene);
+
+  // Box3D's own Large Pyramid benchmark, 20 rows of it: 2870 boxes that stay where they are put
```
Popov72: This three-loop example creates 2,870 boxes, but the upstream Box3D “Large Pyramid” benchmark—and the performance table below—uses a two-dimensional pyramid with 210 boxes at 20 rows. Please either reproduce that two-loop benchmark or call this a separate 3D pyramid so readers do not associate this workload with the published measurements.

Popov72 (changes requested):
The page is valuable and the benchmark data matches the repository results, but the copy/paste setup paths and compatibility wording need correction before merging.

content/communityExtensions/box3dPhysics.md
Pryme8: Added `import "@babylonjs/core/Physics/joinedPhysicsEngineComponent";` to both npm examples (the basic one and the threaded one), plus a paragraph under the first saying what happens without it: on older 9.x `scene.enablePhysics is not a function`, and on 9.26 a side effect stub that does nothing, so the first physics body throws `No Physics Engine available.` instead. Fixed in 4d43122.

content/communityExtensions/box3dPhysics.md
Pryme8: Right, that was a syntax error as written. The setup is now inside an async IIFE. I ran the snippet verbatim as a classic `<script>` against the published CDN files: it parses and enables physics with `Box3DPlugin`. Fixed in 4d43122.

content/communityExtensions/box3dPhysics.md
Pryme8: Agreed, reworded to "supports most of the Physics V2 API", and added a Limitations section that lists the differences that change behaviour: `PhysicsCharacterController`, `MESH`/`HEIGHTFIELD` limited to static and animated bodies, hull and cylinder approximations, how `Physics6DoFConstraint` maps onto Box3D's joints (symmetric swing cone up to 90 degrees, one softness per joint, motors on axes the joint lacks), what `COLLISION_CONTINUED` reports, and that `shapeCast`/`shapeProximity`/`pointProximity` have no equivalent. Each one is checked against the plugin source. Fixed in 4d43122.

content/communityExtensions/box3dPhysics.md
Pryme8: Good catch, it was a 3D pyramid under a benchmark label. The example now builds the benchmark's own scene: the 2D pyramid, 210 unit boxes at 20 rows, with the same positions, ground slab, friction, restitution and box mass (1 tonne, density 1000). Running it verbatim gives a maximum drift of 0.027 m after 5 s, the same figure the benchmark reports for that scene. Fixed in 4d43122.

Popov72 (approved):

bjsplat:

Snapshot: [https://snapshots-cvgtc2eugrd3cgfd.z01.azurefd.net/DOCS/refs/pull/1629/merge](https://snapshots-cvgtc2eugrd3cgfd.z01.azurefd.net/DOCS/refs/pull/1629/merge)
Path: `DOCS/refs/pull/1629/merge`
Build: https://dev.azure.com/babylonjs/ContinousIntegration/_build/results?buildId=59843

